PR #8, opened 2018-04-26
Split Project Into Modules, Add Scribe Support

### LINKED ISSUE #1: Consider using scribe backend
Hey @ChristopherDavenport, this is a really nice project! Looks great!

I was wondering whether you'd consider a PR to use [outr/scribe](https://github.com/outr/scribe) rather than log4s or making the logging backend configurable? scribe has support for scala-native / scala.js and great performance, and it can still be used as an slf4j backend.

ChristopherDavenport:
Breaks Project Out - Multiple Backend Support

Log4s got own project with creators operating from the Log4sLogger object.
Scribe got its own project

Fixes #1


scribe/src/main/scala/io/chrisdavenport/log4cats/scribe/ScribeLogger.scala
```diff
@@ -0,0 +1,44 @@
...
+import io.chrisdavenport.log4cats.Logger
+import scribe.{Logger => Base}
+import cats.effect.Sync
+
+object ScribeLogger {
+
+  def fromLogger[F[_]: Sync](logger: Base): Logger[F] = new Logger[F]{
+    // Law Abiding But Not What We Want
+    def isTraceEnabled: F[Boolean] = 
+      Sync[F].delay(???)
```
ChristopherDavenport: @DavidGregory084 Is there any ability for this functionality in scribe somewhere I'm missing. We may need to build a multi-level algebra for universality is scribe can't support these operations.

scribe/src/main/scala/io/chrisdavenport/log4cats/scribe/ScribeLogger.scala
```diff
@@ -0,0 +1,55 @@
...
+import cats.effect.Sync
+
+object ScribeLogger {
+
+  def fromLogger[F[_]: Sync](logger: Base): Logger[F] = new Logger[F]{
+
+    // This May Be a Very Bad Approach...
+    def isTraceEnabled: F[Boolean] = 
+      Sync[F].delay(
+        logger.modifiers.find(mod => mod(LogRecord.simple("", "", "", None, None, Level.Trace)).isDefined).isDefined
```
ChristopherDavenport: Still likely a terrible idea, but at least has the chance to be succesful.

DavidGregory084: I think we need to look at the `logger.handlers`. I feel like there should be a better way to write this than:
```scala
logger.handlers.exists {
  _.minimumLevel
    .map(_ <= Level.Trace)
    .getOrElse(true)
}.getOrElse(false)
```
or:
```scala
(for {
  handler <- logger.handlers
  minLevel <- handler.minimumLevel.toList
  if minLevel <= Level.Trace
} yield ()).nonEmpty
```
but my brain's too blasted at the moment :laughing:

ChristopherDavenport: `minimumLevel` does not appear to be a value of the handler, only in the creation of the handler.

PR #22, opened 2018-05-17
Add Slf4j-api direct logger implementation

### LINKED ISSUE #17: Direct SLF4J backend
For projects which might use log4cats as their only logging interface, the additional layer of abstraction that is log4s seems unnecessary, would it be possible to build a direct SLF4J backend to log4cats?

ChristopherDavenport:
A direct slf4j-api backend for log4cats. No reflection or anything fancy.

Fixes #17 

Feedback appreciated @keirlawson @rossabaker


ChristopherDavenport:
Going to roll this out as a first draft. Run some benches etc. And then determine whether or not additional work is required.

PR #28, opened 2018-05-22
Mima function improvement

ChristopherDavenport:
Universal MiMa Version Check With SemVer


ChristopherDavenport:
Going to open through other route.

PR #33, opened 2018-05-29
Alternative : Alternative io log4s proposal

ChristopherDavenport:
Alternative possible solution for mdc/awareness/logger algebras.

Creates an alternative implementation that is created in an internal modules and then exposed higher.

@lorandszakacs RFC

### DIFF
--- a/slf4j-internal/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/LoggerMacros.scala
+++ b/slf4j-internal/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/LoggerMacros.scala
@@ -2,7 +2,7 @@
  * Copyright 2013-2017 Sarah Gerweck
  * see: https://github.com/Log4s/log4s
  *
- * Modifications copyright (C) 2018 Lorand Szakacs
+ * Modifications copyright (C) 2018 Christopher Davenport
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


slf4j-internal/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/LoggerMacros.scala
```diff
@@ -0,0 +1,208 @@
...
+    import c.universe._
+
+    val logger = q"${c.prefix.tree}.logger"
+    val F = q"${c.prefix}.F"
+    val logValues = error match {
+      case None => List(msg)
+      case Some(e) => List(msg, e.tree)
+    }
+    val logExpr = q"$logger.${TermName(logLevel.methodName)}(..$logValues)"
+    val checkExpr = q"$logger.${TermName(s"is${logLevel.name}Enabled")}"
```
lorandszakacs: This calls `.isEnabled` on the java logger. I didn't delay this in an `F[_]` because from the looks of it it seemed to be referentially transparent once the logger is created. Creation of the logger itself being the impure operation.

But I wasn't 100% sure 😊

ChristopherDavenport: By what criteria is the creation of the logger not referentially transparent? As in because it loads resources. In that case do you think the methods should return `F[Logger[F]]`

lorandszakacs: @ChristopherDavenport, yes I was thinking about the fact that logger creation loads resources, therefore it is unsafe. 

We could leave current methods as they are (since I have a hunch that even all us FP people still use unsafe creation methods), and maybe add additional methods of the form:`safe${factoryMethod} : F[Logger[F]]`.

This latter approach is slightly problematic since the convention is usually the other way around: adding the `unsafe` word to method names. Therefore, the second option would be to stick to convention, and basically ruin all current factory methods, and provide the same functionality under `unsafe${factoryMethod}: Logger[F]` method names.

I'm also fine with not doing anything else 🤷‍♂️. I have never had problems with the impurity of logger creation, only with the impurity of logging itself 😄.

ChristopherDavenport: I agree. I think I'll throw these into an unsafe name, and with a corresponding equivalent one in `F` just so that we are totally transparent.

PR #39, opened 2018-06-04
Scalaz Log4s Module

ChristopherDavenport:
There was no core algebra linked to cats-effect so creates Scalaz module. So I made this as a proof of concept. Uncertain if this makes sense within log4cats but I don't see a conclusive reason not to. Or I can make it a fully outside repository Which relies on core, and remove the cats/cats-effect from core here. 

RFC's welcome


lorandszakacs:
A great idea! We can keep it in the same repository until our algebra evolves a bit more, and maybe then split it off, if ever.

ChristopherDavenport:
Alright reception has been pretty light so I'll get this in tomorrow.

Issue #44, opened 2018-06-20
Message and Error Logger base classes


ChristopherDavenport:
I've been considerring this and not sure if its a great use of time and resources, as most wouldn't utilize it but a message logger and a error logger split would allow a context logger to become a message logger given a context. So you can set context along the way or other tools to pass a specific context. 

Wondering if others have thought as it puts another split in the hierarchy in a far more fundamental position.

PR #48, opened 2018-06-23
Structure Hierarchy and Structured Logger Modification Function

ChristopherDavenport:
Addresses a way of dealing with #44, so now we have a concrete solution.

### DIFF
--- a/core/shared/src/main/scala/io/chrisdavenport/log4cats/StructuredLogger.scala
+++ b/core/shared/src/main/scala/io/chrisdavenport/log4cats/StructuredLogger.scala
@@ -10,4 +10,13 @@ trait StructuredLogger[F[_]] extends Logger[F] {
 
 object StructuredLogger{
   def apply[F[_]](implicit ev: StructuredLogger[F]): StructuredLogger[F] = ev
+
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/io/chrisdavenport/log4cats/StructuredLogger.scala
```diff
@@ -10,4 +10,13 @@ trait StructuredLogger[F[_]] extends Logger[F] {
 
 object StructuredLogger{
   def apply[F[_]](implicit ev: StructuredLogger[F]): StructuredLogger[F] = ev
+
+  def withContext[F[_]](sl: StructuredLogger[F])(ctx: (String, String)*): MessageLogger[F] =
```
ChristopherDavenport: Is this function worth this shift? Even if primarily invisible to the user.

lorandszakacs:
I like this proposal, and it opens up new avenues of combining stuff. For instance: why couldn't (wouldn't) we have a similar `StructuredLogger`  for errors as well?

ChristopherDavenport:
Alright, I think after we merge this it may be appropriate to start with some binary compatible versions as presently this seems sufficiently stable for most purposes.

PR #92, opened 2018-10-23
Fix Macros

### LINKED ISSUE #91: Slf4jLogger.create[F] doesn't reflect the enclosing class name
```scala
import cats.effect.{ExitCode, IO, IOApp, Sync}
import io.chrisdavenport.log4cats.slf4j.Slf4jLogger

object MyClass {
  def mkLogger[F[_]: Sync] = Slf4jLogger.create[F]
  def mkLoggerUnsafe[F[_]: Sync] = Slf4jLogger.unsafeCreate[F]
}

object LoggerApp extends IOApp {
  override def run(args: List[String]): IO[ExitCode] = for {
    safeLogger <- MyClass.mkLogger[IO]
    unsafeLogger = MyClass.mkLoggerUnsafe[IO]
    _ <- safeLogger.info("safe")
    _ <- unsafeLogger.info("unsafe")
  } yield ExitCode.Success
}
```

Result:
```
INFO  i.c.l.s.Slf4jLogger - safe 
INFO  MyClass - unsafe 
```

Expected:
```
INFO  MyClass - safe  
INFO  MyClass - unsafe 
```

ChristopherDavenport:
Just duplicate the first macro but add in a call to delay in the macro expansion.

I have no idea how to reuse or build composable macros so any insights appreciated.

Fixes #91

Confirmed Fix in another project as its easier than building a test for it. If anyone has a clean way to test that would be appreciated.

### DIFF
--- a/cats/slf4j-internal/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/LoggerMacros.scala
+++ b/cats/slf4j-internal/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/LoggerMacros.scala
@@ -18,7 +18,7 @@
  */
 package io.chrisdavenport.log4cats.slf4j.internal
 
-import scala.annotation.tailrec
+import scala.annotation.{tailrec, StaticAnnotation}
 import scala.reflect.macros.{blackbox, whitebox}
 
 /** Macros that support the logging system.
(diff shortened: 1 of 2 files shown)


ChristopherDavenport:
As this works and fixes a bug, think it makes sense to move forward with this unless anyone objects shortly.

catostrophe:
I've found a better solution. Will publish a bit later today.

ChristopherDavenport:
#94 Seems the more optimal solution

PR #94, opened 2018-10-24
Fix Slf4jLogger.create

catostrophe:
Fix `create` macros without code duplication and static annotations via a bundle.

I also added a few compilation tests. They don't check the name of the logger, but only validate quasiquotes.


catostrophe:
This is to fix #91

Issue #80, opened 2018-10-05
Additional Data Type Instances


ChristopherDavenport:
```
mapK: (F ~> G) => Alg[F] => Alg[G]
Kleisli
OptionT
EitherT
WriterT
Writer: Could be a pure cats backend
```

bpholt:
Once a release is cut that includes #460, I believe this issue can be closed. Could we make such a release?

rossabaker:
Since you know the secret Big Ten handshake, https://github.com/typelevel/log4cats/releases/tag/v1.4.0.  Should be there within an hour or so.  I'll do a 2.x release after #518.

PR #120, opened 2019-01-15
Add WriterT logger

DennisVDB:
Here's an implementation of a `WriterT` logger (see #80). I based myself on how the current `Writer` logger is implemented.

I went through a handful of iterations and I'm still not sure if it's correct (it compiles). Especially, `run: WriterT[F, G[LogMessage], ?] ~> F` seems too constrained. I tried implementation it with the shape `WriterT[F, G[LogMessage], ?] ~> H` but couldn't.

### DIFF
--- a/cats/extras/shared/src/main/scala/io/chrisdavenport/log4cats/extras/WriterTLogger.scala
+++ b/cats/extras/shared/src/main/scala/io/chrisdavenport/log4cats/extras/WriterTLogger.scala
@@ -5,6 +5,13 @@ import cats.data._
 import cats.implicits._
 import io.chrisdavenport.log4cats._
 
+/**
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
As mentioned on gitter my holdout on this is that we either should put very large warning labels on it, or not include it. `WriterT` suffers that if any errors occur in `F` all values written are lost. Which makes sense for the abstraction but can be very surprising. 

I'm inclined to the large warning label documentation, and then not include it in any supplied docs so that its available if people are certain thats what they want and go looking for it, but we sufficiently keep it away from wide use. It certainly has a place.

DennisVDB:
The documentation I've added should straight to the point. What would you think about putting it under `...log4cats.extras.unsafe`?

PR #124, opened 2019-01-23
Internal Macro Improvements

ChristopherDavenport:


ChristopherDavenport:
@lorandszakacs  @catostrophe If anyone can tell me where I've gone off the rails here I'd really appreciate it.

cats/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/LoggerMacros.scala
```diff
@@ -154,67 +157,85 @@ private[slf4j] object ReflectiveLogMacros {
...
+        // val Seq = q"scala.collection.Seq"
         val backup = TermName(c.freshName("mdcBackup"))
         q"""
            if ($checkExpr) $F.delay {
              val $backup = $MDC.getCopyOfContextMap
              try {
-               for ((k, v) <- $Seq(..$context)) $MDC.put(k, v)
+               for {
+                 (k, v) <- $ctxExp.toSeq
+                } yield $MDC.put(k, v)
```
sbuzzard: Do you want the yield here? Desugars to flatmap and map versus foreach without it and since this is done for put side effect only ... I’ve got bitten by weird optimizations around ignored non unit expressions ... likely not related but my two cents from the peanut gallery. :-)

som-snytt: I agree that is not right. Also, converting from `Map` to `Seq` is not necessary for iteration.

som-snytt:
From the travis failure, turning on `-explaintypes` might help. The `Nil.head` might be no type params in call to checkBounds.

cats/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/Slf4jLogger.scala
```diff
@@ -19,62 +19,57 @@
...
 
-  def fromSlf4j[F[_]: Sync](logger: JLogger): F[SelfAwareStructuredLogger[F]] =
-    Sync[F].delay(unsafeFromSlf4j(logger))
+  def getLogger[F[_]: Sync]: SelfAwareStructuredLogger[F] =
+    macro GetLoggerMacros.unsafeCreateImpl[F[_]]
 
+  @deprecated("0.3.0", "Use getLogger instead")
   def unsafeCreate[F[_]: Sync]: SelfAwareStructuredLogger[F] =
-    macro GetLoggerMacros.unsafeCreateImpl[F[_]]
+    getLogger[F]
```
catostrophe: Replace with
 ```scala
macro GetLoggerMacros.unsafeCreateImpl[F[_]]
```
to avoid referencing a macro generated method.

cats/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/LoggerMacros.scala
catostrophe: :+1: for `foreach` here

ChristopherDavenport: Updated back to foreach

PR #139, opened 2019-02-17
Factor out some repetition in slf4j logger internals

tbrown1979:
Also changed the `try/finally` to use a `.guarantee`. Let


tbrown1979:
@ChristopherDavenport Finally got around to changing this back to `try/finally`

PR #188, opened 2019-07-11
Fix issue 187 — MDC — rewrite finalizer w/ F.guarantee

### LINKED ISSUE #187: MDC regression between 0.3.0 and 0.4.0-M1
For a simple setup like this:
build.sbt
```
libraryDependencies += "io.chrisdavenport" %% "log4cats-slf4j" % "0.4.0-M1" withSources()
libraryDependencies += "ch.qos.logback" % "logback-classic" % "1.2.3" withSources()
```

Log4CatsMDC.scala:
```scala
package bug

import cats._
import cats.implicits._
import cats.effect._
import io.chrisdavenport.log4cats.slf4j._

object Log4CatsMDC extends IOApp {
  private val logger = Slf4jLogger.getLogger[IO]
  override def run(args: List[String]): IO[ExitCode] = {
    for {
      _ <- logger.info(Map("p1" -> "value"))("w/ MDC")
      _ <- logger.info("w/o MDC")
    } yield ExitCode.Success
  }
}
```

logback.xml:
```
<configuration>

    <appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
        <encoder>
            <pattern>[%thread] %highlight(%-5level) %cyan(%logger) - %mdc — %msg %n
            </pattern>
        </encoder>
    </appender>

    <root level="DEBUG">
        <appender-ref ref="CONSOLE"/>
    </root>
</configuration>
```
With no other changes except log4cats version, we get:

✅ `0.3.0` 
```
[scala-execution-context-global-479] INFO  bug.Log4CatsMDC - p1=value — w/ MDC 
[scala-execution-context-global-479] INFO  bug.Log4CatsMDC -  — w/o MDC 
```

❌ `0.4.0-M1`
```
[ioapp-compute-0] INFO  bug.Log4CatsMDC -  — w/ MDC 
[ioapp-compute-0] INFO  bug.Log4CatsMDC -  — w/o MDC
```

As you can see, MDC key/value pairs no longer show up in logs 😢

lorandszakacs:
Closes #187 

thanks @oleg-py for pointing out the problem. Indeed it was that.

I assume that the commit that introduced the bug was the one that removed an `F.suspend` from somewhere at some point. Can't recall exactly. Mind might be playing tricks.

### DIFF
--- a/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
+++ b/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
@@ -26,15 +26,16 @@ private[slf4j] object Slf4jLoggerInternal {
     isEnabled.ifM(
       F.suspend {
         val backup = MDC.getCopyOfContextMap
-        try {
+        val putMDC = F.delay {
           for {
             (k, v) <- ctx
           } MDC.put(k, v)
-          logging
-        } finally {
+        }
+        val clearMDC = F.delay {
           if (backup eq null) MDC.clear()
           else MDC.setContextMap(backup)
         }
+        F.guarantee(putMDC *> logging)(clearMDC)
       },
       F.unit
     )


Daenyth:
It might be good to add a test that exposes this as well

PR #227, opened 2019-09-22
Remove dependency on github4s

lavrov:
`github4s` is not used directly in this project.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -151,7 +151,6 @@ lazy val micrositeSettings = Seq(
     "-Ywarn-unused:imports",
     "-Xlint:-missing-interpolator,_"
   ),
-  libraryDependencies += "com.47deg" %% "github4s" % "0.20.1",
   micrositePushSiteWith := GitHub4s,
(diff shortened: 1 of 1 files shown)


ChristopherDavenport:
While this may compile, it will no longer work to publish the microsite.

ChristopherDavenport:
Close was by accident, my apologies.

lavrov:
Dependency trace:
`sbt-microsites` -> [`org-policies-core`](https://github.com/47deg/sbt-microsites/blob/master/project/ProjectPlugin.scala#L41) -> [`github4s`](https://github.com/47deg/sbt-org-policies/blob/master/project/ProjectPlugin.scala#L75)

lavrov:
> While this may compile, it will no longer work to publish the microsite.

If it compiles then `github4s` is not used in that module. Why do you think it won't publish? Sbt project has its own dependencies which are used to perform build related stuff. Can we make an experiment and try? ;)

lavrov:
Just published myself. It works.

ChristopherDavenport:
Alright, we're opting out of new upgrades.  👍

PR #228, opened 2019-09-22
LoggerFactory implementation

### LINKED ISSUE #224: Support LoggerFactory
Abstract `LoggerFactory[F[_]]` which would provide implementation agnostic factory methods. Having such factory would allow one to choose implementation at one place once.

lavrov:
Closes #224

### DIFF
--- a/core/shared/src/main/scala/io/chrisdavenport/log4cats/LoggerFactory.scala
+++ b/core/shared/src/main/scala/io/chrisdavenport/log4cats/LoggerFactory.scala
@@ -1,0 +1,9 @@
+package io.chrisdavenport.log4cats
+
+trait LoggerFactory[F[_]] {
+
+  def fromName(name: String): F[SelfAwareStructuredLogger[F]]
+
+  def fromClass(clazz: Class[_]): F[SelfAwareStructuredLogger[F]]
+
+}
--- a/core/shared/src/main/scala/io/chrisdavenport/log4cats/LoggerFactory.scala
+++ b/core/shared/src/main/scala/io/chrisdavenport/log4cats/LoggerFactory.scala
@@ -5,5 +5,9 @@ trait LoggerFactory[F[_]] {
   def fromName(name: String): F[SelfAwareStructuredLogger[F]]
 
   def fromClass(clazz: Class[_]): F[SelfAwareStructuredLogger[F]]
+}
 
+object LoggerFactory {
+
+  def apply[F[_]](implicit ev: LoggerFactory[F]): LoggerFactory[F] = ev
 }
--- a/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/Slf4jLoggerFactory.scala
+++ b/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/Slf4jLoggerFactory.scala
@@ -3,15 +3,16 @@ package io.chrisdavenport.log4cats.slf4j
 import cats.effect.Sync
 import io.chrisdavenport.log4cats.{LoggerFactory, SelfAwareStructuredLogger}
 
-class Slf4jLoggerFactory[F[_]: Sync] extends LoggerFactory[F] {
-
-  override def fromName(name: String): F[SelfAwareStructuredLogger[F]] =
-    Slf4jLogger.fromName(name)
-
-  override def fromClass(clazz: Class[_]): F[SelfAwareStructuredLogger[F]] =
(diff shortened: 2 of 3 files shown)


slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/Slf4jLoggerFactory.scala
```diff
@@ -0,0 +1,17 @@
+package io.chrisdavenport.log4cats.slf4j
+
+import cats.effect.Sync
+import io.chrisdavenport.log4cats.{LoggerFactory, SelfAwareStructuredLogger}
+
+class Slf4jLoggerFactory[F[_]: Sync] extends LoggerFactory[F] {
```
ChristopherDavenport: If we could put this in the companion object and make it private then we don't need to let anything besides a LoggerFactory interface escape.

core/shared/src/main/scala/io/chrisdavenport/log4cats/LoggerFactory.scala
```diff
@@ -0,0 +1,13 @@
+package io.chrisdavenport.log4cats
+
+trait LoggerFactory[F[_]] {
```
lorandszakacs: We also need methods that do not return in a logger in `F[_]`. Probably in a separate trait, in which we express like stuff [getLogger: SelfAwareStructuredLogger](https://github.com/ChristopherDavenport/log4cats/blob/master/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/Slf4jLogger.scala#L29).

And your `Slf4jLoggerFactory` would implement both.

sirmax: +1 for adding impure but pragmatic `getXxx(…): Logger[F]` methods.

I disagree with adding them in a separate trait though. The biggest virtue of `LoggerFactory[F]` is that the whole application and libraries as well may depend on a common factory to spawn their loggers. And if there would be two or more traits, one would have to make a tough choice for the whole app which trait to use or, alternatively, pass both factories bloating the code.

lorandszakacs: The reason I opt for a separate trait is that sometimes it is impossible to write a factory that returns things in non-F in a safe way. For instance loggers that require some internal buffer that then send stuff somewhere remote — like to AWS cloudwatch or something. You'd need to instantiate some mutable state inside an `F[_]` upon logger creation.

What we can do, maybe, is add a kitchen-sync included factory that inherits from both? Although this might cause the diamond problem in implicit resolution. But if your codebase does not have such stateful loggers, then you can just use the super-trait everywhere and no harm done. But if you do have statefull loggers, then you cannot use the super-trait everywhere.

sirmax: > loggers that require some internal buffer that then send stuff somewhere remote

That's an interesting point. To put this idea even further perhaps `F[Logger[F]]` would not be enough here and `Resource[F, Logger[F]]` would be more appropriate.

I wonder if it's possible to move the state and other heavy machinery insider the factory itself? We would then have something like this
```scala
object CloudwatchLoggerFactory {
  def resource[F: …](…): Resource[F, LoggerFactory[F]] = …
}
``` 
while its `getLogger`s would just set a logger name and thus could be implemented trivially.

_Disclaimer: I'm a bit inexperienced with remote logging, so all of the above might make no sense._

lorandszakacs: I implemented [something for Cloudwatch logging](https://github.com/busymachines/pureharm-aws/blob/master/aws-logger/src/main/scala/busymachines/pureharm/aws/logger/AWSLoggerFactory.scala), and currently only the `LoggerFactory` creation itself has to be a `Resource`. I didn't get to implement the local buffering yet, that's why my factory has non-F methods, but I think that a simple `F[_]` is enough.

lorandszakacs: But that doesn't mean that `Resource[F *]` won't be necessary in other cases 😄

mr-git: @lavrov, https://gitter.im/ChristopherDavenport/libraries?at=5e5413111ec15e02119ae192

PR #230, opened 2019-09-24
Add ApplicativeLocal and FunctorTell loggers

iRevive:
This PR adds two new loggers:
1) A logger that uses a value of `cast.mtl.ApplicativeLocal` as an additional MDC;
2) A logger that writes `LogMessage` entries into effect context using `cats.mtl.FunctorTell`;

These changes related to the https://github.com/ChristopherDavenport/log4cats/issues/143.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,7 @@
 import sbtcrossproject.{crossProject, CrossType}
 val catsV = "2.0.0"
 val catsEffectV = "2.0.0"
+val catsMtlV = "0.7.0"
 val slf4jV = "1.7.28"
 val specs2V = "4.7.1"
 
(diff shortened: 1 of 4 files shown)


ChristopherDavenport:
My only concern with having this in the main project is that cats-mtl is still pre-1.0 and we are post-1.0. Would you mind if I created a project for this that we included prominently that could be sub 1.0 versioned? Otherwise we risk a major on a change to mtl which makes me nervous.

iRevive:
@ChristopherDavenport I am completely fine with a separated project until the major release of cats-mtl.

ChristopherDavenport:
Here is where I have the baseline of the project setup. https://github.com/ChristopherDavenport/log4cats-mtl

I can either move your work over, or a PR over there would be appreciated for us to get this rolling. The deploy is all setup, so it should just be a matter of getting the code in place.

iRevive:
I will create a PR

iRevive:
Done https://github.com/ChristopherDavenport/log4cats-mtl/pull/1

PR #281, opened 2020-02-04
Add the StructuredTestingLogger

ollyw:
Following up on the Gitter chat in December, this PR adds a StructuredTestingLogger that provides a way to easily test structured logging. It is fairly simple. There is one point that might merit discussion. The log message case classes have a default for the ctx parameter such as:
```
  final case class TRACE(message: String, throwOpt: Option[Throwable], ctx: Map[String, String] = Map.empty) extends LogMessage
```
This means that in classes where the context is missing in many cases, it makes it easier to test without explicitly providing Map.empty each time. The option is there to also provide a default for the throwOpt, so the tests only have to provide the mandatory message. This has the advantage that is to easier to read, but the disadvantage that it is 1) less explict, 2) Inconsistent with the TestLogger which doesn't have a default parameter. The test code would look kinda like this (ignoring the unsafe!):
```
      logger.logged.unsafeRunSync() mustBe Vector(
        INFO("msg1"),
        WARN("msg2", ctx = Map("foo" -> "bar")),
        ERROR("msg3", throwOpt = Some(MyException("test ex")))
      )
```

Anyway, let me know your thoughts.
Thanks!

### DIFF
--- a/testing/shared/src/main/scala/io/chrisdavenport/log4cats/testing/StructuredTestingLogger.scala
+++ b/testing/shared/src/main/scala/io/chrisdavenport/log4cats/testing/StructuredTestingLogger.scala
@@ -1,0 +1,106 @@
+package io.chrisdavenport.log4cats.testing
+
+import io.chrisdavenport.log4cats.SelfAwareStructuredLogger
+import cats.effect.Sync
+import cats.implicits._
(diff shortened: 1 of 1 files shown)


ollyw:
Thanks for approving. Is it clear when this change may be put into a release?

PR #298, opened 2020-03-26
Make MDC loging safe in the presence of cancelation

### LINKED ISSUE #274: MDC is not cleared on task cancellation
In the case where the logging task (with a context for MDC) gets cancelled, the MDC context is not cleared/restored to it's previous state.

Here is an [example code](https://github.com/ikempf/log4cats-mdc-clear/blob/master/src/main/scala/com/ikempf/Start.scala) that quickly produces the issue. A log setting the MDC context gets cancelled. Other log tasks then randomly display keys from the "leaked" MDC context. 

I've had a look at the [MDC related code](https://github.com/ChristopherDavenport/log4cats/blob/master/slf4j/src/main/scala/io/chrisdavenport/log4cats/slf4j/internal/Slf4jLoggerInternal.scala). I was about to suggest using a `Resource` or a `Bracket` to clear/restore the MDC to it's previous state but that's already the case.
I'm not sure why this is not working as expected 🤔, I'll have another look this weekend.

kubukoz:
Closes #274


kubukoz:
I think I'll redo it without SyncIO...

kubukoz:
I'll think of a test, shouldn't be too hard...

kubukoz:
Okay, I added a test that ensures that the cancelation runs on a different thread (while we're logging) and that both threads get the correct MDC context after the whole thing. The old version fails the test.

kubukoz:
All good on my side :)

kubukoz:
@ChristopherDavenport would you mind releasing? I'd like to depend on this behavior :)

ChristopherDavenport:
https://travis-ci.org/github/ChristopherDavenport/log4cats/builds/685830390 - I may need to transition onto github actions which I haven'tn done yet

kubukoz:
I'm getting that error on master too, will submit a fix :)

kubukoz:
https://github.com/ChristopherDavenport/log4cats/pull/313

PR #343, opened 2020-09-20
Add a `LogLevel.fromString` method

Daenyth:
I have this in my codebase, figured it makes sense to push upstream.

Do we want it to match the output of `show` at all?


lorandszakacs:
@Daenyth 

I would prefer that we match the output of `show` with that of `fromString`. 

And for the sake of not breaking the existing API I think it's best to adapt your method. Otherwise I wouldn't care either way.

Also, just remembered I'm a maintainer of this repo 😬

Daenyth:
That's reasonable. Feel free to edit this branch before merging if I don't get to it first

PR #347, opened 2020-10-07
Add implicits for EitherT and OptionT transformers

tdbgamer:
@ChristopherDavenport Let me know if there are any issues with this. Library is great so far, but I end up making these instances in my projects so I figured it'd make more sense to contribute it here if you're okay with it.

### DIFF
--- a/core/shared/src/main/scala/io/chrisdavenport/log4cats/Logger.scala
+++ b/core/shared/src/main/scala/io/chrisdavenport/log4cats/Logger.scala
@@ -1,6 +1,7 @@
 package io.chrisdavenport.log4cats
 
 import cats._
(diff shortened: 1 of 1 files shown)


rossabaker (approved):
I'm new here, but these look reasonable to me.

tdbgamer:
Awesome, so is there anything else I need to do before merge?

rossabaker:
I was being cautious about merging to see if another maintainer wanted to weigh in, but I don't think these should be controversial.

I'm going to be working on releases for the move to Typelevel, with full Dotty support, and a CE3 upgrade on a second branch.  I don't have rights to deploy a patch version under Chris' org before The Big move.

PR #361, opened 2020-11-26
Add withModifiedContext to StructuredLogger

ivan-klass:
Add ability to modify context, similar to `Logger.withModifiedString`
Example: mask some sensitive structure field values
```scala

val currentLogger: StructuredLogger[F] = ???
val maskRe = "(?<=.{3}).(?=[^@]*?@)".r
val hideEmailLogger = StructuredLogger.withModifiedContext(currentLogger) {
  ctx => ctx.get("clientEmail").fold(ctx)(
    email => ctx.updated("clientEmail", maskRe.replaceAllIn(email, "*"))
  )
}
```

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
@@ -32,7 +32,7 @@ object SelfAwareStructuredLogger {
     new ModifiedContextSelfAwareStructuredLogger[F](sl)(ctx ++ _)
 
   private class ModifiedContextSelfAwareStructuredLogger[F[_]](sl: SelfAwareStructuredLogger[F])(
-    modify: Map[String, String] => Map[String, String]
+      modify: Map[String, String] => Map[String, String]
   ) extends SelfAwareStructuredLogger[F] {
     private lazy val defaultCtx: Map[String, String] = modify(Map.empty)
     def error(message: => String): F[Unit] = sl.error(defaultCtx)(message)
--- a/core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
@@ -29,27 +29,27 @@ object SelfAwareStructuredLogger {
   def withContext[F[_]](
       sl: SelfAwareStructuredLogger[F]
   )(ctx: Map[String, String]): SelfAwareStructuredLogger[F] =
-    new ExtraContextSelfAwareStructuredLogger[F](sl)(ctx)
+    new ModifiedContextSelfAwareStructuredLogger[F](sl)(ctx ++ _)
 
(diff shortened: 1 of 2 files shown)


rossabaker:
I'm just trying to do releases. Paging the experts who use this library more extensively before I merge.

core/shared/src/main/scala/io/chrisdavenport/log4cats/StructuredLogger.scala
```diff
@@ -21,49 +21,54 @@ object StructuredLogger {
...
+  private class ModifiedContextStructuredLogger[F[_]](sl: StructuredLogger[F])(
+      modify: Endo[Map[String, String]]
   ) extends StructuredLogger[F] {
-    private val outer = ctx
-    def error(message: => String): F[Unit] = sl.error(outer)(message)
-    def warn(message: => String): F[Unit] = sl.warn(outer)(message)
-    def info(message: => String): F[Unit] = sl.info(outer)(message)
-    def debug(message: => String): F[Unit] = sl.debug(outer)(message)
-    def trace(message: => String): F[Unit] = sl.trace(outer)(message)
+    private def whenEmpty: Map[String, String] = modify(Map.empty)
```
lorandszakacs, line 34: This can be easily turned back into a `val`, right? Since applying a pure function to the same value, yields the same result.

core/shared/src/main/scala/io/chrisdavenport/log4cats/StructuredLogger.scala
```diff
@@ -21,49 +21,54 @@ object StructuredLogger {
...
 
-  private class ExtraContextStructuredLogger[F[_]](sl: StructuredLogger[F])(
-      ctx: Map[String, String]
+  def withModifiedContext[F[_]](
+      sl: StructuredLogger[F]
+  )(modifyCtx: Endo[Map[String, String]]): StructuredLogger[F] =
+    new ModifiedContextStructuredLogger[F](sl)(modifyCtx)
+
+  private class ModifiedContextStructuredLogger[F[_]](sl: StructuredLogger[F])(
+      modify: Endo[Map[String, String]]
```
lorandszakacs, line 32: While `Endo` is really cool, I think it might trip up more people than necessary. I think the classic function syntax would be more approachable.

lorandszakacs (changes requested):
I like this addition to the library :+1: 

I would like to see the same method added to `org.typelevel.log4cats.SelfAwareStructuredLogger` as well to maintain predictability in usage.

core/shared/src/main/scala/io/chrisdavenport/log4cats/StructuredLogger.scala
ivan-klass: @lorandszakacs thanks! I've replaced with `lazy val`

ivan-klass:
> I would like to see the same method added to `org.typelevel.log4cats.SelfAwareStructuredLogger` as well to maintain predictability in usage.
Added similar method to `SelfAwareStructuredLogger`

lorandszakacs (approved):

lorandszakacs:
LGTM :+1: 

@ChristopherDavenport, do you want to review this?

rossabaker (approved):
:+1: We'll need to not forget to merge this forward to main before releasing that.

PR #356, opened 2020-11-08
Use GitHub Actions and accross the board updates

cquiroz:
Use Github Actions instead of travis to build and includes a bunch of library updates

PR #364, opened 2020-12-04
use GitHub actions

jopecko:


rossabaker:
Thanks! #356 did this and more, but you used the plugin.  I'm going to work on merging that one t get the "and more", and then reconcile it with this one.

rossabaker:
This is still going to fail pending #370, but is closer to where we want to be.

larsrh:
`master` works now (except for the release, because nobody set that up)

PR #370, opened 2021-01-10
Fix warnings

rossabaker:


rossabaker:
The build is broken due to deprecations in the GitHub Actions.

PR #373, opened 2021-01-16
Move to Typelevel

rossabaker:


build.sbt
```diff
@@ -12,6 +12,14 @@ ThisBuild / scalaVersion := Scala213
 ThisBuild / publishFullName := "Christopher Davenport"
 ThisBuild / publishGithubUser := "christopherdavenport"
 
+ThisBuild / versionIntroduced := Map(
+  // First versions after the Typelevel move
+  "2.12" -> "1.1.2",
+  "2.13" -> "1.1.2",
+  "3.0.0-M2" -> "1.1.2",
+  "3.0.0-M3" -> "1.1.2",
```
rossabaker, line 20: This is a lie.  Why doesn't "1.2.0" work here?

.github/workflows/ci.yml
```diff
@@ -12,10 +12,13 @@ on:
     branches: ['*']
   push:
     branches: ['*']
-    tags: [v*]
+    tags: [v*, v*]
```
ChristopherDavenport, line 15: Why twice?

ChristopherDavenport (approved):

.github/workflows/ci.yml
rossabaker: It's a quirk in sbt-github-actions that I haven't managed to ask about.

PR #374, opened 2021-01-16
Dotty

rossabaker:
Compiles, but the tests are sparse, and I have no faith that the logger names are correct.  We're using the meaty macro out of log4s.


rossabaker:
Uh, do these builds have a history of hanging?

ChristopherDavenport:
Nope, not generally

ChristopherDavenport (commented):
This looks good, just need to fix up the whatever's hanging.

rossabaker:
It's not just Dotty, which is weird.  Not so much else changed.

rossabaker:
If it hangs again, hopefully the dotty-hang branch will expose it.

rossabaker:
It's caught in the mess of `dirtyStuff`, but I don't know why it's broken now.

rossabaker:
So, non-deterministically, getting the MDC variable for the finalizer thread was happening from the finalizer thread.  I gave it a third thread to that we can shift back to as "main", ensuring that the producer is not the consumer on the blocking calls, and it passed on all builds.  I will run again for greater confidence.

lorandszakacs (approved):
LGTM. Did a quick and dirty verification of the logger names.

Will add some tests for it in the near future.

rossabaker:
Thanks, @lorandszakacs! There's a nice [test suite in log4s](https://github.com/Log4s/log4s/blob/master/core/shared/src/test/scala/org/log4s/GetLoggerSpec.scala) that might be something we could adapt to get some rigor on the logger names.

rossabaker:
Thanks, @lorandszakacs! There's a nice [test suite in log4s](https://github.com/Log4s/log4s/blob/master/core/shared/src/test/scala/org/log4s/GetLoggerSpec.scala) that might be something we could adapt to get some rigor on the logger names.

rossabaker:
Thanks, @lorandszakacs! There's a nice [test suite in log4s](https://github.com/Log4s/log4s/blob/master/core/shared/src/test/scala/org/log4s/GetLoggerSpec.scala) that might be something we could adapt to get some rigor on the logger names.

ChristopherDavenport (approved):

PR #382, opened 2021-01-20
Cats-Effect 3 support

rossabaker:


slf4j/src/test/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternalSuite.scala
```diff
@@ -100,30 +87,29 @@ class Slf4jLoggerInternalSuite extends CatsEffectSuite {
...
-        runBlockingOn(MDC.put(variable, initial))(loggerThread)
-        runBlockingOn(MDC.put(variable, initial))(finalizerThread)
-      } *>
+    (IO {
+      IO(MDC.put(variable, initial)).evalOn(loggerThread)
+      IO(MDC.put(variable, initial)).evalOn(finalizerThread)
+    } *>
       performLogging
-        .start(IO.contextShift(loggerThread))
+        .evalOn(loggerThread)
```
rossabaker, line 104: This doesn't run the fiber on `loggerThread` like I hoped. What's up?

rossabaker:
Logging is still `Sync`, right?

slf4j/src/test/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternalSuite.scala
RaasAhsan: So I inserted a `println(Thread.currentThread().getName())` on `performLogging` and I see `my-thread-1` which seems to be the expected case.

I started looking into why the test times out, and it's basically a deadlock: `logCanceled.await()` is blocking the thread and so the fiber is never released. When `performCancel` tries to cancel the fiber, that's going to block forever because the `performLogging` fiber never suspended itself. I tried sticking all those blocking calls into an `IO.blocking` block but that didn't seem to work, possibly for different reasons.

rossabaker: I got tricked: the log line happening on iocompute is from the other test.

The same basic test works in cats-effect-2.  The `fiber.cancel` returns immediately, which counts down `logCanceled`.  (The canceled log does render.)  And the exit case of the `performLogging`  task is indeed Canceled.

```scala
    val performLogging = Slf4jLogger
      .getLogger[IO]
      .info(Map(variable -> "modified")) {
        startedLog.countDown()
        println("AWAITING")
        logCanceled.await()
        println("AWAITED")        
        finishedLog.countDown()
        "Heavy to compute value you're logging"
      }.guaranteeCase(kase => IO(println(kase)))

    def performCancel[A](fiber: Fiber[IO, A]): IO[Unit] = {
      IO(startedLog.await()) *>
      IO(println("CANCELING")) *>
        fiber.cancel *>
      IO(println("CANCELED")) *>        
        IO(logCanceled.countDown())
    }
```

Something has changed in fiber cancellation.

RaasAhsan: Yeah, so CE2 cancellation was immediate -- the canceller runs the finalizers of the target fiber, which could be unsafe in some circumstances. In CE3, there is a notion of "owning the runloop" now. In order to cancel, you need to own the runloop, and that's basically implemented as a lock.

When you run a blocking piece of code in CE3 in just a regular `delay` node, like `latch.await()`, there is no chance that a canceller will be able to cancel it; the runloop was never released. `IO.blocking` is intended for these kinds of situations. I tried it really quickly and tests were still hanging, but maybe you'll find more success with it :)

rossabaker: I was hoping this would yield, but it succeeds after 10 seconds rather than being canceled:

```
    val performLogging = Slf4jLogger
      .getLogger[IO]
      .info(Map(variable -> "modified")) {
        startedLog.countDown()
        (1 to 100).map { _ =>
          IO.blocking(logCanceled.await(100, TimeUnit.MILLISECONDS)).unsafeRunSync()
        }
        finishedLog.countDown()
        "Heavy to compute value you're logging"
      }
```

I think the essence of the test is a delayed block that finishes, but only through cancellation.

rossabaker: I can easily cancel things we indefinitely block on inside this delay, but I haven't found a way to cancel the delayed block itself.  I'm not convinced that this test still makes sense.

RaasAhsan: If you block forever inside a `delay`, it can never be canceled. I think I would be more helpful here if I spent some time to understand the test

rossabaker: I didn't write the test, but my interpretation of it is to ensure that the MDC is reset if the logging is canceled. The logging can't be cancled, so I think there is nothing here to test anymore.

ChristopherDavenport:
Only way it wouldn't be is if we cheated. It's definitely executing a side-effect.

rossabaker:
Somewhere in some library someone smuggled what I thought was clearly a Sync into a Concurrent, so now I'm questioning everything I thought I knew whenever I do CE3 work. I wish I remembered where I got this idea.

djspiewak:
I *believe* that if you swap every `IO.apply` for an `IO.interruptible(false)`, the test will start working. The flow dependency here is definitely super-wonky and relies on the unsound CE2 semantics. In CE3, neither `delay` *nor* `blocking` is cancelable, and they will hold the runloop until their execution completes (this is, to my knowledge, the only safe semantic). `interruptible` is basically a variant of `blocking` which *is* cancelable, but only provided that the thunk within `interruptible` is within a thread suspension (so, something that's actually interruptible via `Thread#interrupt`).

rossabaker:
I don't think it's probably worthwhile or necessary here: is anybody doing interruptible work to calculate a log string?  Stuff we could, or would even want, to cancel?

djspiewak:
> I don't think it's probably worthwhile or necessary here: is anybody doing interruptible work to calculate a log string? Stuff we could, or would even want, to cancel?

Very, very unlikely.

m-sp (approved):

ChristopherDavenport (approved):

PR #425, opened 2021-04-10
Add fluent combinators for adding context to structured loggers

### LINKED ISSUE #424: Streamline adding context to structured loggers
At the moment, to add some context to the logger we have to do this:
```scala
StructuredLogger.withContext(logger)(Map("foo" -> 123.show))
// Or
SelfAwareStructuredLogger.withContext(logger)(Map("foo" -> 123.show))
```
This is noisy. Also note that the two versions produce subtly different results - the first one "loses" the logger's self-awareness. 

Instead, I'd like to be able to write something like
```scala
logger.withContext(Map("foo" -> "123"))
// or, better yet
logger.withContext("foo" -> "123"))
```
and have it return `StructuredLogger` or `SelfAwareStructuredLogger` depending on the `logger`'s type.

nigredo-tori:
Closes #424

- `addContext(Map[String, String])` members are trivial wrappers over existing `withContext` functions. I couldn't call this `withContext` because for some reason that disabled static forwarders for `StructuredLogger.withContext` etc., which broke binary compatibility. :shrug: Thankfully, I like the new name better than the old one, because it makes it clear that the context is appended rather than replaced.
- `addContext((String, Shown)*)` is sugar over it using the `Show.Shows` magnet. Consider:
    ```scala
    logger.addContext("foo" -> 123)
    // versus
    logger.addContext(Map("foo" -> 123.show))
    ```

We can add similar sugar for `error`, `info` etc. - but I'm not sure those would be useful enough to justify the required explosion of the number of methods.


lorandszakacs:
I do like the idea, and it has my OK. But before that:

1) the feature should branch off of `series/1.x` branch, and we forward port it to `main` series 2.x
2) the method should exist on `StructuredLogger[F[_]]`, and then override it in `SelfAwareStructuredLogger[F[_]]` as well. The reason for is analogous to the problem fixed in [this PR](https://github.com/typelevel/log4cats/pull/430)

nigredo-tori:
> 1. the feature should branch off of `series/1.x` branch, and we forward port it to `main` series 2.x

I've rebased this PR to 1.x, with no conflicts.

> 2. the method should exist on `StructuredLogger[F[_]]`, and then override it in `SelfAwareStructuredLogger[F[_]]` as well. The problem is analogous to the `withModifiedString` method, fixed in #430

Could you please elaborate? As far as I see, this is already OK - both methods are introduced in `StructuredLogger[F[_]]` (returning `StructuredLogger[F]`), and both are explicitly overriden in `SelfAwareStructuredLogger[F[_]]` (returning `SelfAwareStructuredLogger[F]`).

lorandszakacs:
> Could you please elaborate? As far as I see, this is already OK - both methods are introduced in `StructuredLogger[F[_]]` (returning `StructuredLogger[F]`), and both are explicitly overriden in `SelfAwareStructuredLogger[F[_]]` (returning `SelfAwareStructuredLogger[F]`).

You're right, sorry. Must have overlooked :sweat_smile: all good :+1:

ChristopherDavenport (approved):

PR #429, opened 2021-04-16
Backport sbt 1.5.0 upgrade, and plugins cleanup

lorandszakacs:
Series 1.x is not dead yet, so just doing maintenance.


lorandszakacs:
Not entirely sure why this fails in CI, because `githubWorkflowsCheck` works fine locally :sweat:

lorandszakacs:
Ok, figured it out, initially did not backport the removal of the `csbt` runner.

PR #446, opened 2021-05-14
Scala 3.0.0 support

lorandszakacs:
Added `sbt-dotty` until this is merged and released: https://github.com/djspiewak/sbt-spiewak/pull/54

Other than that it's quite straightforward. Dropping RC Scala version support.

After this is merged, I'll handle forward porting this to main as well 😄 

LE:
We can't publish without `sbt-dotty` `0.5.5` because `0.5.3` (pulled in by `sbt-spiewak`) sets the binary version for Scala 3 to `_3.0.0` instead of the required `_3`


build.sbt
```diff
@@ -3,12 +3,13 @@ import sbtghactions.UseRef
...
 val Scala213 = "2.13.5"
 val Scala212 = "2.12.13"
+val Scala3   = "3.0.0"
 
 enablePlugins(SonatypeCiReleasePlugin)
 
 ThisBuild / organization := "org.typelevel"
 ThisBuild / baseVersion := "1.3"
-ThisBuild / crossScalaVersions := Seq(Scala213, Scala212, "3.0.0-RC2", "3.0.0-RC3")
+ThisBuild / crossScalaVersions := Seq(Scala3, Scala213, Scala212)
```
rossabaker, line 12: Does reordering these make Scala 3 the default version?  If so, does that irritate tooling?

build.sbt
```diff
@@ -21,7 +22,8 @@ ThisBuild / versionIntroduced := Map(
   "3.0.0-M3" -> "1.2.0",
   "3.0.0-RC1" -> "1.2.0",
   "3.0.0-RC2" -> "1.2.2",
-  "3.0.0-RC3" -> "1.3.0"
+  "3.0.0-RC3" -> "1.3.0",
+  "3.0.0" -> "1.3.1",
```
rossabaker, line 26: I think we can clean out all the old `RC` values here.

rossabaker (approved):
Looks great. A couple optional comments.

build.sbt
lorandszakacs: Did a quick double check crossScalaVersions only influences the order in which `++` prefixed tasks set the version which to run. 

And the default Scala version when you open up sbt  is set by the `scalaVersion` property.

I guess it makes sense for the first one in x-compilation to be the same  as `scalaVersion`

rossabaker: Oh, that's right.  I'm probably thinking of some org plugin I've used somewhere that infers it from the cross versions.

PR #518, opened 2021-11-05
Add PagingSelfAwareStructuredLogger

dj707chen:
Add PagingSelfAwareStructuredLogger.

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2018 Christopher Davenport
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.typelevel.log4cats
+
+import cats._
+import cats.implicits._
+
+private class PagingSelfAwareStructuredLogger[F[_]: Monad](pageSize: Int)(
+    sl: SelfAwareStructuredLogger[F]
+) extends SelfAwareStructuredLogger[F] {
+  private def paging(
+      loggingLevelChk: => F[Boolean],
+      loggingOp: String => F[Unit],
+      msg: => String
+  ): F[Unit] = {
+    loggingLevelChk.ifM(
+      {
+        val length = msg.getBytes().length
+        val decorationSize =
+          Math.max(
+            Math.min(pageSize / 10, 2048),
+            512
+          )
+        val pageContentSize = pageSize - decorationSize
+        if (length <= pageContentSize)
+          loggingOp(msg)
+        else {
+          val pl = msg.grouped(pageContentSize).toList.zip(LazyList from 1)
+          val numOfPages = pl.length
+          pl.traverse(msgIndexTuple =>
+            loggingOp(
+              show"""
+                  |Page ${msgIndexTuple._2} / $numOfPages
+                  |${msgIndexTuple._1}""".stripMargin
+            )
+          )
+        }
+      },
+      Applicative[F].unit
+    )
+  }
+
+  override def trace(ctx: Map[String, String])(msg: => String): F[Unit] =
+    paging(isTraceEnabled, sl.trace(ctx), msg)
+
+  override def trace(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+    paging(isTraceEnabled, sl.trace(ctx, t), msg)
+
+  override def debug(ctx: Map[String, String])(msg: => String): F[Unit] =
+    paging(isDebugEnabled, sl.debug(ctx), msg)
+
+  override def debug(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+    paging(isDebugEnabled, sl.debug(ctx, t), msg)
+
+  override def info(ctx: Map[String, String])(msg: => String): F[Unit] =
+    paging(isInfoEnabled, sl.info(ctx), msg)
+
+  override def info(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+    paging(isInfoEnabled, sl.info(ctx, t), msg)
+
+  override def warn(ctx: Map[String, String])(msg: => String): F[Unit] =
+    paging(isWarnEnabled, sl.warn(ctx), msg)
+
+  override def warn(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+    paging(isWarnEnabled, sl.warn(ctx, t), msg)
+
+  override def error(ctx: Map[String, String])(msg: => String): F[Unit] =
+    paging(isErrorEnabled, sl.error(ctx), msg)
+
+  override def error(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+    paging(isErrorEnabled, sl.error(ctx, t), msg)
+
+  override def isTraceEnabled: F[Boolean] = sl.isTraceEnabled
+
+  override def isDebugEnabled: F[Boolean] = sl.isDebugEnabled
+
+  override def isInfoEnabled: F[Boolean] = sl.isInfoEnabled
+
+  override def isWarnEnabled: F[Boolean] = sl.isWarnEnabled
+
+  override def isErrorEnabled: F[Boolean] = sl.isErrorEnabled
+
+  override def error(message: => String): F[Unit] =
+    paging(isErrorEnabled, sl.error, message)
+
+  override def warn(message: => String): F[Unit] =
+    paging(isWarnEnabled, sl.warn, message)
+
+  override def info(message: => String): F[Unit] =
+    paging(isInfoEnabled, sl.info, message)
+
+  override def debug(message: => String): F[Unit] =
+    paging(isDebugEnabled, sl.debug, message)
+
+  override def trace(message: => String): F[Unit] =
+    paging(isTraceEnabled, sl.trace, message)
+
+  override def error(t: Throwable)(message: => String): F[Unit] =
+    paging(isErrorEnabled, sl.error(t), message)
+
+  override def warn(t: Throwable)(message: => String): F[Unit] =
+    paging(isWarnEnabled, sl.warn(t), message)
+
+  override def info(t: Throwable)(message: => String): F[Unit] =
+    paging(isInfoEnabled, sl.info(t), message)
+
+  override def debug(t: Throwable)(message: => String): F[Unit] =
+    paging(isDebugEnabled, sl.debug(t), message)
+
+  override def trace(t: Throwable)(message: => String): F[Unit] =
+    paging(isTraceEnabled, sl.trace(t), message)
+}
--- a/core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
@@ -40,6 +40,11 @@ trait SelfAwareStructuredLogger[F[_]] extends SelfAwareLogger[F] with Structured
 object SelfAwareStructuredLogger {
   def apply[F[_]](implicit ev: SelfAwareStructuredLogger[F]): SelfAwareStructuredLogger[F] = ev
 
+  def withPaging[F[_]: Monad](pageSize: Int)(
+      logger: SelfAwareStructuredLogger[F]
+  ): SelfAwareStructuredLogger[F] =
+    new PagingSelfAwareStructuredLogger(pageSize)(logger)
+
   def withContext[F[_]](
       sl: SelfAwareStructuredLogger[F]
   )(ctx: Map[String, String]): SelfAwareStructuredLogger[F] =


core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,149 @@
...
+      loggingLevelChk.ifM(
+        {
+          val pageSize = pageSizeK * 1000
+          val length = msg.getBytes().length
+          if (length <= pageSize)
+            loggingOp(msg)
+          else {
+            val decorationSize = 256
+            val pageContentSize = pageSize - decorationSize
+            val pageList = msg.grouped(pageContentSize).toList
```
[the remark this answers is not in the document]
rossabaker: It's also going to allocate and hold all the substrings at once, instead of one substring at a time. This isn't normally a big deal, but if the log message is large enough to page, it might be large enough to matter.  Instead of `.grouped` out here, you might just calculate each chunk from the pageIndex inside the `traverse_` with a bit of `substring` math.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,149 @@
...
+
+    private def pagedLogging(
+        loggingLevelChk: => F[Boolean],
+        loggingOp: (=> String) => F[Unit],
+        msg: => String
+    ): F[Unit] = {
+      loggingLevelChk.ifM(
+        {
+          val pageSize = pageSizeK * 1000
+          val length = msg.getBytes().length
```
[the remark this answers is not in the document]
rossabaker: Actually, you're chopping this up by characters, not bytes.  Multibyte characters don't "weigh" any more in your grouping.  Why not just `msg.length`?

rossabaker (changes requested):
Thanks!  It's a nice idea.  Some ideas below.

You might also be able to write some unit tests with the `StructuredTestingLogger`.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,149 @@
...
+  /**
+   * Wrap a SelfAwareStructuredLogger adding pagination functionality.
+   *
+   * @param pageSizeK The size (unit is kilobyte) of the chunk of message in each page; this does not include the
+   *                  page header and footer, and tracing data.
+   * @param logger The SelfAwareStructuredLogger to be used to do the actual logging.
+   * @tparam F Effect type class.
+   * @return SelfAwareStructuredLogger with pagination.
+   */
+  def withPaging[F[_]: Monad](pageSizeK: Int, maxPageNeeded: Int)(
```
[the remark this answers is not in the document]
dj707chen: Sorry Ross, I was busy on other issues and could not respond to your comments earlier. The motivation of using K as unit of page size is to make it easier to determine the value, the reason we want to do paging is because the underline logging system has some size limit on the a single log entry, some will split a large log entry into multiple entries, but cut off the logging context, by setting the page size below that size limit you can avoid splitting by logging system; usually the log size limit should be in the order of KB.
Adding scaladoc for  maxPageNeeded.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,149 @@
...
+    private val pageIndices = 1 to maxPageNeeded
+
+    private def pagedLogging(
+        loggingLevelChk: => F[Boolean],
+        loggingOp: (=> String) => F[Unit],
+        msg: => String
+    ): F[Unit] = {
+      loggingLevelChk.ifM(
+        {
+          val pageSize = pageSizeK * 1000
```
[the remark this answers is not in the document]
dj707chen: Per the purpose of this whole thing is to avoid splitting by underline logging provider, we want to be in the safe range, in case the user is assuming 1000 for K.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,149 @@
...
+          val length = msg.getBytes().length
+          if (length <= pageSize)
+            loggingOp(msg)
+          else {
+            val decorationSize = 256
```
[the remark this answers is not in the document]
dj707chen: For the page header and footer. I will remove it, but add scaladoc comment to the pageSizeK parameter that this should be taken into consideration when deciding page size in addition to the log size limit of underline logging provider.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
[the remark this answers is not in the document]
dj707chen: Thanks for the insight, will just do that.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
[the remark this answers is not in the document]
dj707chen: Will add unit tests.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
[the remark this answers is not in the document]
dj707chen: Thanks for the suggestion, will do that.

build.sbt
```diff
@@ -119,7 +120,9 @@ lazy val testing = crossProject(JSPlatform, JVMPlatform)
   .settings(
     name := "log4cats-testing",
     libraryDependencies ++= Seq(
-      "org.typelevel" %%% "cats-effect" % catsEffectV
+      "org.typelevel" %%% "cats-effect"     % catsEffectV,
+      "ch.qos.logback"  % "logback-classic" % logbackClassicV % Test,
+      "org.specs2"     %% "specs2-core"     % specs2V         % Test
```
rossabaker, line 125: I'd prefer to keep the tests in MUnit rather than introduce a second framework, but we can clean that up in a subsequent PR.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,223 @@
...
+   * Example: Assume the log size limit of the underline logging provider is 75 Kb, setting pageSizeK to 64 leaves
+   * 11 Kb for (part 2) and (part 3).
+   *
+   * @param pageSizeK The size (in unit kibibyte) of the chunk of message in each page.
+   * @param maxPageNeeded The maximum number of pages to be logged.
+   * @param logger The SelfAwareStructuredLogger to be used to do the actual logging.
+   * @tparam F Effect type class.
+   * @return SelfAwareStructuredLogger with pagination.
+   */
+  @SuppressWarnings(Array("org.wartremover.warts.DefaultArguments"))
```
rossabaker, line 50: We're not using wartremover on this project, are we?

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,223 @@
...
+        msg: => String
+    ): F[Unit] = {
+      val numOfPagesRaw = (msg.length - 1) / pageSize + 1
+      val numOfPages = Math.min(numOfPagesRaw, maxPageNeeded)
+      if (numOfPages <= 1)
+        loggingOp(msg)
+      else {
+        val logSplitIdPart1 = logSplitId.split('-').head
+        val pageHeaderTail = s"$numOfPages $logSplitIdPart1"
+        val pageFooterTail = s"$numOfPages $logSplitIdN=$logSplitId page_size=$pageSizeK Kb"
```
rossabaker, line 78: `KiB` is a less ambiguous abbreviation when you're using kibibytes.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,223 @@
...
+  ): SelfAwareStructuredLogger[F] =
+    new PagingSelfAwareStructuredLogger[F](pageSizeK, maxPageNeeded)(logger)
+
+  private class PagingSelfAwareStructuredLogger[F[_]: Monad](pageSizeK: Int, maxPageNeeded: Int)(
+      sl: SelfAwareStructuredLogger[F]
+  ) extends SelfAwareStructuredLogger[F] {
+    assert(pageSizeK > 0)
+    assert(maxPageNeeded > 0)
+
+    private val pageIndices = (1 to maxPageNeeded).toList
```
rossabaker, line 62: I'm not sure whether `Range` has a `.traverse_` instance, but see if it compiles without the `.toList` to make it lazy.  If you can't, no big deal, but you're making a list of 999 numbers that you won't usually use.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,223 @@
...
+    ): F[Unit] = {
+      loggingLevelChk.ifM(
+        doLogging(loggingLevelChk, logOpWithCtx, s"$msg\n${getStackTrace(t)}", ctx),
+        Applicative[F].unit
+      )
+    }
+
+    def getStackTrace(t: Throwable): String = {
+      val bos = new ByteArrayOutputStream()
+      val ps = new PrintStream(bos)
```
rossabaker, line 140: I think you either want to flush this stream or pass `true` as a second argument to this constructor.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,223 @@
...
+      t.printStackTrace(ps)
+      bos.toString
```
rossabaker, line 142: This depends on the system's default character set.  You can circumvent all those nasty character encoding issues by using `Writer`s instead of `Stream`s.  Here's the commons-lang implementation in Java.  It's _almost_ the same as yours.

```java
   public static String getStackTrace(final Throwable throwable) {
        final StringWriter sw = new StringWriter();
        final PrintWriter pw = new PrintWriter(sw, true);
        throwable.printStackTrace(pw);
        return sw.getBuffer().toString();
    }
```

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,223 @@
...
+  def withPaging[F[_]: Monad](pageSizeK: Int = 64, maxPageNeeded: Int = 999)(
+      logger: SelfAwareStructuredLogger[F]
+  ): SelfAwareStructuredLogger[F] =
+    new PagingSelfAwareStructuredLogger[F](pageSizeK, maxPageNeeded)(logger)
+
+  private class PagingSelfAwareStructuredLogger[F[_]: Monad](pageSizeK: Int, maxPageNeeded: Int)(
+      sl: SelfAwareStructuredLogger[F]
+  ) extends SelfAwareStructuredLogger[F] {
+    assert(pageSizeK > 0)
+    assert(maxPageNeeded > 0)
```
rossabaker, line 60: These should be recoverable errors for an application.  I'd throw `IllegalArgumentException` instead of `assert`, which raises an error that won't be caught by `NonFatal`.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -0,0 +1,223 @@
...
+        loggingOp: (=> String) => F[Unit],
+        logSplitId: String,
+        msg: => String
+    ): F[Unit] = {
+      val numOfPagesRaw = (msg.length - 1) / pageSize + 1
+      val numOfPages = Math.min(numOfPagesRaw, maxPageNeeded)
+      if (numOfPages <= 1)
+        loggingOp(msg)
+      else {
+        val logSplitIdPart1 = logSplitId.split('-').head
```
rossabaker, line 76: This will throw if `logSplitId` is ever exclusively `-` characters.  I'm not sure if that's possible.

rossabaker (commented):
Looking good!  A few more ideas below.

build.sbt
dj707chen: Will do that later

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
dj707chen: No, removing it.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
dj707chen: Good suggestion, thanks.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
dj707chen: No luck. So I am creating a much smaller List of the size of numOfPages instead of 999 then taking from it.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
dj707chen: Good suggestion, thanks.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
dj707chen: Took it, thanks.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
dj707chen: It's created at Line 95 as a UUID, it's safe to do that.
```
val logSplitId = UUID.randomUUID().show
```

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
dj707chen: Good suggestion, took it.

build.sbt
dj707chen: Done

testing/shared/src/test/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLoggerTestRunner.scala
```diff
@@ -48,49 +48,52 @@ class PagingSelfAwareStructuredLoggerTestRunner extends CatsEffectSuite {
...
+            }
+            assertEquals(allMsgValid, true)
           }
-
-          ctxValid && msgValid
-        }
-        assertEquals(allMsgValid, true)
-      }
+        )
+      )
```
rossabaker: Getting rid of the `unsafeRunSync` in tests is good, but I don't think the `unsafeRunAsync` is quite what we want.  Assertion failures are thrown by the async callback, but not intercepted and recorded by the test framework.  When I run this with false assertions, it hangs.  We want to compose these assertions into an IO and then let MUnit run them.

rossabaker (commented):
Main code is looking good.  I have one remaining concern with effa481.

rossabaker (approved):
:+1: if the build is green

danicheg:
Seems like we can have this for the `series/1.x` as well. Do we want it?

PR #568, opened 2021-12-28
Add syntax for the `Logger`

danicheg:
This adds syntax that allows logging in that way:
```scala
def log[F[_]: Sync: Logger] = 
  for {
    result1 <- successComputation[F]
    _ <- info"First result is $result1"
    _ <- errorComputation[F].onError(_ => error"We got an error!")
  } yield ()
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,3 +55,25 @@ object MyThing {
   } yield something
 }
 ```
+
+### Laconic syntax
+
+It's possible to use interpolated syntax for logging.
+Currently, supported ops are: `trace`, `debug`, `info`, `warn`, `error`.
+You can use it for your custom `Logger` as well as for Slf4j `Logger`.
+
+```scala mdoc
+import cats.effect.Sync
+import org.typelevel.log4cats.Logger
+import org.typelevel.log4cats.syntax._
+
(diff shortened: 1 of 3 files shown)


rossabaker (commented):
Clever, but I'm not sure that I'd use it.  Open to what others think.

danicheg:
@ChristopherDavenport would you check this, please?

danicheg:
No one else hasn't expressed an opinion about this 😞 I'd be happy if can to decide on our own.

rossabaker (approved):
I prefer to refer to lexically-scoped loggers that log to a known category, rather than pass them as constraints, where I'm not sure what category they'll log to.  Nor do I usually care to track logging as a capability.  So I can't imagine myself using this.

However, you like it, it doesn't pollute the namespace without an import, and it doesn't bring in any dependencies.  It's good for you and it won't hurt me.  I say your :+1: weighs more than my :-1:.

danicheg:
Thanks! Personally, I'm 50/50 about using logging via capabilities. But there is well known in one coterie logging library (tofu-logging), which promotes this approach pretty much. Therefore thinking about users familiar with that approach, it could be nice to have it in our's arsenal.

PR #581, opened 2022-01-08
Set `organizationName` and regenerate headers

danicheg:


ChristopherDavenport (approved):
I think this needs the end year for apache, since we're regenerating.

rossabaker (approved):
I went down a rabbit hole on end years a while back and found some incompatible opinions:

1. You artificially shorten your copyright if you don't update.
2. Setting it without materially changing the file is fraud.
3. Each header should contain the set of years that file was changed.
4. Your copyright already extends N years past the death of the last contributor, where N might equal 70.
5. There are international treaties that cover most of the world, but probably not all of us.
6. In a few decades, nobody is going to give a shit anyway.

I find 4 through 6 the most compelling, and fortuitously, low hassle.  No more end dates for me.

danicheg:
Ross's arguments look convincing to me. And I'm happy to merge this as is.

PR #587, opened 2022-01-14
Fixes for log4cats-1.5

rossabaker:
The build was creating headers, but not checking headers.  This resulted in a diff, which bumped the version past the tag and resulted in a resolution failure on MiMa.  Cool.

I took a conservative approach and did not port to sbt-typelevel.


danicheg (approved):
Thanks! My bad that I didn't check the header in #568 😞

rossabaker:
No, I would have assumed CI took care of it.  This is why I'm excited for sbt-typelevel and reducing all these exceptional builds.

PR #590, opened 2022-01-14
Update slf4j-api to 1.7.33 on series/1.x

rossabaker:


rossabaker:
Ope, this one came in while I was asleep.

PR #592, opened 2022-01-19
Migrate to sbt-typelevel

armanbilge:
Yeah yeah I know, this needs a backport to the other branch as well.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,84 +1,28 @@
-import sbtghactions.UseRef
-
 val Scala213 = "2.13.8"
 val Scala212 = "2.12.15"
 val Scala3 = "3.0.2"
 
-enablePlugins(SonatypeCiReleasePlugin)
-
-ThisBuild / organization := "org.typelevel"
-ThisBuild / organizationName := "Typelevel"
-ThisBuild / baseVersion := "2.2"
+ThisBuild / tlBaseVersion := "2.2"
 ThisBuild / crossScalaVersions := Seq(Scala213, Scala212, Scala3)
 ThisBuild / scalaVersion := Scala213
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -88,65 +32,43 @@ val logbackClassicV = "1.2.10"
...
     name := "log4cats-noop"
   )
-lazy val noopJVM = noop.jvm
-lazy val noopJS = noop.js
 
 lazy val slf4j = project
-  .settings(commonSettings, releaseSettings)
-  .dependsOn(coreJVM)
+  .settings(commonSettings)
+  .dependsOn(core.js)
```
danicheg, line 71: Not a blocker at all, just curious, why did you change the `coreJVM` to `core.js` here?

danicheg (approved):

build.sbt
```diff
@@ -88,65 +32,43 @@ val logbackClassicV = "1.2.10"
...
-  .settings(commonSettings, releaseSettings)
+  .settings(commonSettings)
   .settings(
     name := "log4cats-core",
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-core" % catsV
     )
   )
-lazy val coreJVM = core.jvm
-lazy val coreJS = core.js
```
armanbilge, line 123: I deleted `coreJVM` and friends here. They are effectively redundant.

build.sbt
armanbilge: See my comment https://github.com/typelevel/log4cats/pull/592/files#r794407004. It's a style of code I usually see in legacy builds.

danicheg:
I think this is pretty good to go onward. We should also backport this for the branch with CE 2. I could tackle on, but it may take some time.

PR #603, opened 2022-01-31
Migrate to `sbt-typelevel` in `series/1.x`

danicheg:
This is a backport of #592.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,83 +1,31 @@
-import sbtghactions.UseRef
-
 val Scala213 = "2.13.8"
 val Scala212 = "2.12.15"
 val Scala3 = "3.0.2"
 
-enablePlugins(SonatypeCiReleasePlugin)
-
-ThisBuild / organization := "org.typelevel"
-ThisBuild / organizationName := "Typelevel"
-ThisBuild / baseVersion := "1.5"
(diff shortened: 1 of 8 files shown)


build.sbt
```diff
@@ -6,6 +6,8 @@ ThisBuild / tlBaseVersion := "1.5"
 ThisBuild / crossScalaVersions := Seq(Scala213, Scala212, Scala3)
 ThisBuild / scalaVersion := Scala213
 ThisBuild / startYear := Some(2018)
+ThisBuild / tlSitePublishBranch := None
```
armanbilge, line 9:
```suggestion
```

If you publish the site for 1.x tags, then it will override the 2.x site. Either we should set up versioned documentation, or you should never published the site from the 1.x branch/tags ever. WDYT?

danicheg: Eh, you're probably right. I think sbt-typelevel should have some explicit settings to disable publishing site. It could be odd from some angle but it makes sense in the case of the multibranch project without a versioned site. I think it's worth discussing it in the sbt-typelevel space.

danicheg: @armanbilge I've started https://github.com/typelevel/sbt-typelevel/issues/137, so would be lovely to get your thoughts on it.

PR #622, opened 2022-02-20
Add favicon to the microsite

danicheg:


armanbilge:
Since this is so important to you I can merge https://github.com/typelevel/sbt-typelevel/pull/133 instead of hacking all the builds :) I really wish we could solve https://github.com/typelevel/sbt-typelevel/issues/144.

danicheg:
@armanbilge I think that `sbt-typelevel` should fix that in its own manner, considering the plans about having a dedicated template. Just because it will take some time, I think we may add this temporary workaround for some projects (I'll be the first who will remove it after fixing it in the upstream 😄).

armanbilge:
Ok thanks :) I'm very sorry I didn't prioritize this favicon issue 😬  btw apologies for interfering, I'm just trying really hard to keep all the builds unified as much as possible, I was also complaining in https://github.com/typelevel/discipline/pull/318 😉

danicheg:
This will wait for https://github.com/typelevel/sbt-typelevel/pull/180

Issue #492, opened 2021-09-30
Usability issue with `Sync` constraint requirement.


hejfelix:
The fact that all the constructor methods in `Slf4jLogger` require a `Sync` instance is very detrimental to the usage of this library -- especially when considering that the macro call needs to live inside every class to produce the desired class name in the log lines. This effectively means all classes that log need the `Sync` constraint. 
I'm not sure there is an easy solution here, though :/

FunFunFine:
There is a solution, we can either incapsulate Logger creation ability or delay ability. 

The first option is kind of obvious, here is the simplified example:
 ```scala
trait Slf4jLoggers[F[_]] {
  def make: Slf4jLogger[F]
}
object Slf4jLoggers {
  implicit def fromSync[F[_]: Sync]: Slf4jLoggers[F] = //... delay unsafe loggers creation
}

class MyService[F[_]: Slf4jLoggers] {
  val logger: Logger[F] = Slf4jLogger.getForClassOrSomethingElse
}
```
But this would require us to rework existing code, make some tough design decisions and is not what I would call an easy solution.

The second option can be seen in some libraries already, e.g. tofu: [it's the Delay typeclass,](https://github.com/tofu-tf/tofu/blob/e3551e30f789154e9154ef78f87372c4aeb1d1f9/modules/kernel/src/main/scala/tofu/Delay.scala) which is "weaker" than Sync. Its instance can be derived directly from the Sync instance, but not the other way around.
Then the logger creation would require only `Delay` in scope, so the code would be less detrimental.

@danicheg is it possible to use `tofu.Delay` here or even simply copy-paste it into the project (just like with the laconic syntax)? 
Or perhaps the factory option looks more pleasant, WDYT?

FunFunFine:
I believe this issue can be closed as the #629 was released.

PR #629, opened 2022-03-24
Add LoggerFactory typeclass

hamnis:
The reasoning for this PR is that I want to avoid having a Sync instance just for being able to log in a http4s api.

Extract LoggerName from macro and make it an implicit.
This makes it easier to have forwarders to Slf4jLogger.

This should be binary compatible.


slf4j/src/main/scala-2/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
```diff
@@ -18,13 +18,19 @@ package org.typelevel.log4cats.slf4j
...
+import org.typelevel.log4cats.slf4j.internal.{_}
 import org.slf4j.{Logger => JLogger}
+import scala.annotation.nowarn
 
 object Slf4jLogger {
 
-  def getLogger[F[_]: Sync]: SelfAwareStructuredLogger[F] =
-    macro GetLoggerMacros.unsafeCreateImpl[F[_]]
+  @nowarn("cat=unused")
+  private def getLogger[F[_]: Sync]: SelfAwareStructuredLogger[F] = ???
```
hamnis, line 28: not sure if this is safe to do, please advice

danicheg:
This also resolves #492. I can say that there are uncounted requests on it from TF-folks. So it's nice to see this PR 👍🏻

slf4j/src/main/scala-2/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
Daenyth: Looking again, I think the solution here is to keep the old macro in place, but make this method `private[slf4j]`

Then it's a source-incompatible change, but in practice calling code won't need editing - the same code will just call the new method

slf4j/src/main/scala-2/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
```diff
@@ -18,13 +18,19 @@ package org.typelevel.log4cats.slf4j
...
+  @nowarn("cat=unused")
+  private def create[F[_]: Sync]: F[SelfAwareStructuredLogger[F]] = ???
+
+  def getLogger[F[_]: Sync](implicit name: LoggerName): SelfAwareStructuredLogger[F] =
+    getLoggerFromName(name.value.stripSuffix("$"))
```
FunFunFine, line 33: It seems like the suffix stripping part is the `LoggerName` responsibility. I can't imagine any cases where anyone would need `FooService$`

slf4j/src/main/scala-2/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala
```diff
@@ -0,0 +1,44 @@
...
+  def getLoggerFromClass(clazz: Class[_]): SelfAwareStructuredLogger[F]
+  def getLoggerFromSlf4j(logger: JLogger): SelfAwareStructuredLogger[F]
+  def create(implicit name: LoggerName): F[SelfAwareStructuredLogger[F]]
+  def fromName(name: String): F[SelfAwareStructuredLogger[F]]
+  def fromClass(clazz: Class[_]): F[SelfAwareStructuredLogger[F]]
+  def fromSlf4j(logger: JLogger): F[SelfAwareStructuredLogger[F]]
+}
+
+object Slf4jLoggerFactory {
+  def apply[F[_]](lf: Slf4jLoggerFactory[F]): Slf4jLoggerFactory[F] = lf
```
FunFunFine, line 19:
```suggestion
  def apply[F[_]: Slf4jLoggerFactory]: Slf4jLoggerFactory[F] = implicitly
```

slf4j/src/main/scala-2/org/typelevel/log4cats/slf4j/internal/GetLoggerMacros.scala
```diff
@@ -53,15 +51,6 @@ private[slf4j] class GetLoggerMacros(val c: blackbox.Context) {
...
-    def loggerByParam(param: c.Tree) = {
-      val unsafeCreate =
-        q"_root_.org.typelevel.log4cats.slf4j.Slf4jLogger.getLoggerFromSlf4j(_root_.org.slf4j.LoggerFactory.getLogger(...${List(param)}))($f)"
-      if (delayed)
-        q"_root_.cats.effect.Sync.apply($f).delay(...$unsafeCreate)"
-      else
-        unsafeCreate
-    }
-
     def loggerBySymbolName(s: Symbol) = {
```
FunFunFine, line 54: This method does not return logger anymore, so let's rename it

slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala
```diff
@@ -0,0 +1,48 @@
...
+import org.typelevel.log4cats.SelfAwareStructuredLogger
+import org.typelevel.log4cats.slf4j.internal.*
+import org.slf4j.Logger as JLogger
+import org.typelevel.log4cats.slf4j.Slf4jLogger.{
+  getLoggerFromClass,
+  getLoggerFromName,
+  getLoggerFromSlf4j
+}
+
+trait Slf4jLoggerFactory[F[_]] {
```
FunFunFine, line 13: Is it possible to avoid duplication? This factory trait looks the same for scala 2 and scala 3.

slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala
```diff
@@ -0,0 +1,48 @@
...
+  def getLoggerFromName(name: String): SelfAwareStructuredLogger[F]
+  def getLoggerFromClass(clazz: Class[_]): SelfAwareStructuredLogger[F]
+  def getLoggerFromSlf4j(logger: JLogger): SelfAwareStructuredLogger[F]
+  def create(using name: LoggerName): F[SelfAwareStructuredLogger[F]]
+  def fromName(name: String): F[SelfAwareStructuredLogger[F]]
+  def fromClass(clazz: Class[_]): F[SelfAwareStructuredLogger[F]]
+  def fromSlf4j(logger: JLogger): F[SelfAwareStructuredLogger[F]]
+}
+
+object Slf4jLoggerFactory {
```
FunFunFine, line 24: A little note regarding naming: `cats.effect` uses a way to name factory traits like so:
```scala
trait Ref[F[_]]
//...

object Ref {
  trait Make[F[_]]{
    //...
  } 
}
```
WDYT of using the same naming here for consistency? 






```

Comment (Daenyth) [reply slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala:24]: Those aren't entirely intended for public consumption, not sure if it makes sense to copy that here.

Comment (FunFunFine) [reply slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala:24]: Oh, it seems like I got the idea wrong. Could you please show an example of how to use log4cats with slf4j now?

Comment (FunFunFine) [reply slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala:24]: Isn't it like that?
```scala
class MyService[F[_]: Monad : Slf4jLoggerFactory] {

val logger = Slf4jLogger.getLogger
//methods
}
```

hamnis: The goals is to be able to use log4cats without needing Sync at the callsite. So using it like you describe in https://github.com/typelevel/log4cats/pull/629#discussion_r834472307 correct.

The instance still needs `Sync`.

hamnis: `Slf4jLoggerFactory[F].getLogger` would be used instead

FunFunFine: > `Slf4jLoggerFactory[F].getLogger` would be used instead

It's a bit verbose I think. 
Wouldn't it be more convenient to add summoner method right to the Slf4jLogger object like that:
```scala
object Slf4jLogger {
  def getLogger[F[_]: Sync]] = //same as before
 // same methods as before
 def make[F[_]: Slf4jLoggerFactory]: SelfAwareStructuredLogger[F] = Slf4jLoggerFactory[F].getLogger
}
````
This way API changes kinda less.

Comment (FunFunFine) [reply slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala:24]: > Those aren't entirely intended for public consumption, not sure if it makes sense to copy that here.

You mean `Ref.Make` or `Slf4jLoggerFactory`?

Comment (Daenyth) [reply slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala]: > > Those aren't entirely intended for public consumption, not sure if it makes sense to copy that here.
> 
> You mean `Ref.Make` or `Slf4jLoggerFactory`?

I mean `Ref.Make`

File: slf4j/src/main/scala/org/typelevel/log4cats/slf4j/LoggerFactory.scala [hunk 7]
```diff
@@ -0,0 +1,96 @@
...
+ * limitations under the License.
+ */
+
+package org.typelevel.log4cats
+package slf4j
+
+import cats.effect.Sync
+import org.slf4j.{Logger => JLogger}
+
+trait LoggerFactory[F[_]] {
```
lorandszakacs, line 23: Could this trait be moved to `core`, without the `slf4j` specific methods, and then have a `Slf4jLoggerFactory` that extends it to add `JLogger` methods?

This way, we make log4cats more pallatable for library use, since it doesn't automatically drag in slf4j, and end-users will mostly still instantiate the slf4j specific factory.

It's analogous to what I did in my dead PR. the [core trait](https://github.com/typelevel/log4cats/pull/421/files#diff-188b9221733d2f8a286292dfc94d60f9cbe0d2990b49419ba5b1e0c7af9b8cc4R40), and the [slf4j trait](
https://github.com/typelevel/log4cats/pull/421/files#diff-896fa2ef5da2824bcbf0f8323cdf5f61041b609a3c9006967d9489b01c26409dR25).

hamnis: that is very possible, but I would like some feedback from one of the maintainers for the way forward.

danicheg: I'm not a core contributor of `log4cats`, but I think that's very good advice.

slf4j/src/main/scala-2/org/typelevel/log4cats/slf4j/LoggerNamePlatform.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.typelevel.log4cats.slf4j
+
+import org.typelevel.log4cats.slf4j.internal.GetLoggerMacros
+
+trait LoggerNamePlatform {
```
danicheg, line 21: Basically `Platform` suffix is used when dealing with `JVM`/ `JS`-specific things. Probably `Compat` is more suitable here.

slf4j/src/test/scala/org/typelevel/log4cats/slf4j/LoggerNameTest.scala
```diff
@@ -0,0 +1,37 @@
...
+  object FooB {
+    val name = LoggerName.name
+  }
+
+  test("names") {
+    val name1 = new FooA().name
+    val name2 = FooB.name
+
+    assertEquals(name1.value, "org.typelevel.log4cats.slf4j.LoggerNameTests.FooA")
+    assertEquals(name2.value, "org.typelevel.log4cats.slf4j.LoggerNameTests.FooB")
```
danicheg, line 35: This will fix tests, I guess:
```suggestion
    assertEquals(name1.value, "org.typelevel.log4cats.slf4j.LoggerNameTest.FooA")
    assertEquals(name2.value, "org.typelevel.log4cats.slf4j.LoggerNameTest.FooB")
```

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/LoggerName.scala
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.typelevel.log4cats.slf4j
+
+object LoggerName extends LoggerNamePlatform
+
+final case class LoggerName(value: String)
```
danicheg, line 21: I'm not a real fan of `AnyVal`, but here it seems actually will serve well. What do you think? Or maybe we could just add a type alias for the `String` and use it when needed?

danicheg (commented):
Thanks for your work on this!

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/LoggerName.scala
FunFunFine: Type alias won't work because it is going to be the same as `implicit loggerName: String`, and nobody likes implicit strings

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
```diff
@@ -0,0 +1,82 @@
...
+  def fromName[F[_]: Sync](name: String): F[SelfAwareStructuredLogger[F]] =
+    Sync[F].delay(getLoggerFromName(name))
+
+  def fromClass[F[_]: Sync](clazz: Class[_]): F[SelfAwareStructuredLogger[F]] =
+    Sync[F].delay(getLoggerFromClass(clazz))
+
+  def fromSlf4j[F[_]: Sync](logger: JLogger): F[SelfAwareStructuredLogger[F]] =
+    Sync[F].delay(getLoggerFromSlf4j[F](logger))
+
+  trait Factory[F[_]] extends LoggerFactoryGen[F, SelfAwareStructuredLogger[F]] {
```
hamnis, line 51: not sure if I like this. Perhaps we should rather have a toplevel trait.

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/LoggerName.scala
hamnis: Added the AnyVal suggestion

slf4j/src/main/scala-3/org/typelevel/log4cats/slf4j/Slf4jLoggerFactory.scala
hamnis: `Slf4jLogger.Factory` is what we have now... not sure if I like that.

core/shared/src/main/scala/org/typelevel/log4cats/LoggerFactory.scala
```diff
@@ -22,36 +22,10 @@ import scala.annotation.implicitNotFound
 Implicit not found for LoggerFactory[${F}].
 Information can be found here: https://log4cats.github.io/logging-capability.html
 """)
-trait LoggerFactory[F[_]] extends LoggerFactoryGen[F, SelfAwareStructuredLogger[F]]
-object LoggerFactory extends LoggerFactoryCompanion {
-  def apply[F[_]: LoggerFactory]: LoggerFactory[F] = implicitly
+trait LoggerFactory[F[_]] extends LoggerFactoryGen[F] {
```
hamnis, line 25: we can consider dropping this, since LoggerFactoryGen now has everything that is needed.

FunFunFine: It's kinda convinient for users who would like to use just `LoggerFactory` without the need to define LoggerType explicitly

hamnis:
Any chance this can get merged? @ChristopherDavenport ?

ChristopherDavenport (approved):
This is fully backwards compatible so this would be a minor version right?

hamnis:
That is correct. There are no mima issues or exclusions here

valencik:
This seems awesome, thanks for the hardwork!

I definitely wouldn't want to block this, but is there any chance a bit of documentation could be added, or an example of how one might use this?

lorandszakacs (approved):

ChristopherDavenport:
Going to roll this out tomorrow. Really awesome work.

Issue #643, opened 2022-05-02
log4cats-slf4j depends on log4cats-core_sjs1


martijnhoekstra:
the POM from maven central:

```
<?xml version='1.0' encoding='UTF-8'?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0">
    <modelVersion>4.0.0</modelVersion>
    <groupId>org.typelevel</groupId>
    <artifactId>log4cats-slf4j_2.13</artifactId>
    <packaging>jar</packaging>
    <description>log4cats-slf4j</description>
    <url>https://typelevel.org/log4cats</url>
    <version>2.3.0</version>
    <licenses>
        <license>
            <name>Apache-2.0</name>
            <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
            <distribution>repo</distribution>
        </license>
    </licenses>
    <name>log4cats-slf4j</name>
    <inceptionYear>2018</inceptionYear>
    <organization>
        <name>Typelevel</name>
        <url>https://typelevel.org/log4cats</url>
    </organization>
    <scm>
        <url>https://github.com/typelevel/log4cats</url>
        <connection>scm:git:https://github.com/typelevel/log4cats.git</connection>
        <developerConnection>scm:git:git@github.com:typelevel/log4cats.git</developerConnection>
    </scm>
    <developers>
        <developer>
            <id>christopherdavenport</id>
            <name>Christopher Davenport</name>
            <url>https://christopherdavenport.github.io/</url>
            <email>chris@christopherdavenport.tech</email>
        </developer>
        <developer>
            <id>lorandszakacs</id>
            <name>Loránd Szakács</name>
            <url>https://github.com/lorandszakacs</url>
            <email>lorand.szakacs@protonmail.com</email>
        </developer>
    </developers>
    <properties>
        <info.apiURL>https://www.javadoc.io/doc/org.typelevel/log4cats-slf4j_2.13/2.3.0/</info.apiURL>
        <info.versionScheme>early-semver</info.versionScheme>
    </properties>
    <dependencies>
        <dependency>
            <groupId>org.scala-lang</groupId>
            <artifactId>scala-library</artifactId>
            <version>2.13.8</version>
        </dependency>
        <dependency>
[...]

PR #644, opened 2022-05-03
slf4j depends on core jvm, not core js

martijnhoekstra:
should close https://github.com/typelevel/log4cats/issues/643

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -74,7 +74,7 @@ lazy val noop = crossProject(JSPlatform, JVMPlatform)
 
 lazy val slf4j = project
   .settings(commonSettings)
-  .dependsOn(core.js)
+  .dependsOn(core.jvm)
   .settings(
     name := "log4cats-slf4j",
     libraryDependencies ++= Seq(


lorandszakacs:
I think this would require a 2.4 release? cc @danicheg if you can weigh in please, I got rusty with releases for the past year :sweat_smile:

martijnhoekstra:
The issue causes log4cats-slf4j not to declare its dependency on log4cats-core, and to declare an unneeded dependency on log4cats-core_sjs1

The chance should result in *only* a different POM. That means in terms of binary compatibility, it's backwards and forwards compatible, and doesn't require a minor bump.

build.sbt
```diff
@@ -74,7 +74,7 @@ lazy val noop = crossProject(JSPlatform, JVMPlatform)
 
 lazy val slf4j = project
   .settings(commonSettings)
-  .dependsOn(core.js)
```
danicheg, line 77: Actually, that was discussed when @armanbilge was adding it https://github.com/typelevel/log4cats/pull/592#pullrequestreview-865833002.

armanbilge: Oops. This is 💯  my mistake and I _really_ must not have been paying attention since your comment point it out 😳 I think I was trying to change `coreJVM` to `core.jvm` but I wrote `core.js` ... sorry everyone.

armanbilge:
I agree, this can go into a 2.3.1.

build.sbt
martijnhoekstra: In a Just world, either the build system or the compiler should have given you a fatal warning for this. This is a way too easy mistake to make.

armanbilge: What really baffles me is that the tests run and pass using the JS artifact ...

danicheg: @armanbilge It's my fault too. I wasn't insistent and thought it's no matter what we'd use in that place :)

danicheg:
@martijnhoekstra Could you please rebase your branch? I've fixed the error that caused the failing CI in this PR.

martijnhoekstra:
Yes, rebased

danicheg (approved):
Thank you so much for fixing this!

PR #647, opened 2022-05-09
Merge `series/1.x` into `main`

danicheg:


build.sbt
```diff
@@ -53,6 +53,10 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
       else Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided)
     }
   )
+  .jsSettings(
+    // https://www.scala-js.org/news/2022/04/04/announcing-scalajs-1.10.0#fixes-with-compatibility-concerns
+    libraryDependencies += "org.scala-js" %%% "scalajs-java-securerandom" % "1.0.0"
```
danicheg, line 58: It figured out that `scalajs-java-securerandom` isn't published for Scala 3 https://github.com/scala-js/scala-js-java-securerandom/issues/5. And the `core` module isn't dependent on CE, so using its `Random` isn't an option.

PR #653, opened 2022-05-18
Add LoggerFactory.mapK

alexcardell:
I've been playing around with Natchez and natchez-extras-log4cats. This proved to be a required step to use LoggerFactory rather than instantiating a Kleisli logger and passing around explicitly/implicitly.

I would have applied this to LoggerFactoryGen but with the override type I struggled, I think you would need to lift it into the typeclass signature, but suggestions welcome.

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/LoggerFactory.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/LoggerFactory.scala
@@ -33,17 +33,16 @@ trait LoggerFactory[F[_]] extends LoggerFactoryGen[F] {
 object LoggerFactory extends LoggerFactoryGenCompanion {
   def apply[F[_]: LoggerFactory]: LoggerFactory[F] = implicitly
 
-  def mapK[F[_]: Functor, G[_]](fk: F ~> G)(lfg: LoggerFactory[F]): LoggerFactory[G] =
+  def mapK[F[_]: Functor, G[_]](fk: F ~> G)(lf: LoggerFactory[F]): LoggerFactory[G] =
     new LoggerFactory[G] {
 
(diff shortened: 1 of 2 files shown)


danicheg (commented):
Thanks! One suggestion:

core/shared/src/main/scala/org/typelevel/log4cats/LoggerFactory.scala
```diff
@@ -28,4 +32,18 @@ trait LoggerFactory[F[_]] extends LoggerFactoryGen[F] {
 
 object LoggerFactory extends LoggerFactoryGenCompanion {
   def apply[F[_]: LoggerFactory]: LoggerFactory[F] = implicitly
+
+  def mapK[F[_]: Functor, G[_]](fk: F ~> G)(lfg: LoggerFactory[F]): LoggerFactory[G] =
```
danicheg, line 36: I'd make this method private and add it to the `LoggerFactory` directly (just like in `SelfAwareStructuredLogger`). What do you think?

alexcardell: Yep, good suggestion. Added!

alexcardell:
Maybe slightly redundant but I've added a compilation check like the LoggerSyntaxCompilation

danicheg (approved):
Great job, thanks again.

ChristopherDavenport (approved):

kierendavies:
This is exactly what I need, thank you! When might we see this included in a new release?

danicheg:
Oh, we really didn't release this. I think it'd be fine to do a new minor release `v2.3.2`. Will sort out it.

kierendavies:
Thank you :heart:

PR #667, opened 2022-06-29
Add LoggerFactory instance for NoOpLogger

seigert:
Useful in tests and when your want to disable logging with `LoggerFactory[F]` altogether.

### DIFF
--- a/noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
+++ b/noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2018 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.typelevel.log4cats
+package noop
+
+import cats.Applicative
+
+import scala.annotation.unused
+
+final class NoOpFactory[F[_]](implicit F: Applicative[F]) extends LoggerFactory[F] {
+  def getLoggerFromF: SelfAwareStructuredLogger[F] = NoOpLogger.impl[F]
+
+  def fromF: F[SelfAwareStructuredLogger[F]] = F.pure(getLoggerFromF)
+
+  override def getLoggerFromName(@unused name: String): SelfAwareStructuredLogger[F] =
+    getLoggerFromF
+
+  override def fromName(@unused name: String): F[SelfAwareStructuredLogger[F]] =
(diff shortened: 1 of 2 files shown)


noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.typelevel.log4cats
+package noop
+
+import cats.Applicative
+
+final class NoOpFactory[F[_]](implicit F: Applicative[F]) extends LoggerFactory[F] {
```
danicheg, line 22: It's a good idea to expose a noop `LoggerFactory` instance. Just I'm not sure about having `NoOpFactory` as a separate public abstraction. If we make `NoOpFactory` as a private implementation of `LoggerFactory` and add one public constructor to the companion, it'd be nice.

noop/shared/src/main/scala/org/typelevel/log4cats/noop/package.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.typelevel.log4cats
+
+import cats.Applicative
+
+package object noop {
+  implicit def loggerFactoryForApplicative[F[_]: Applicative]: NoOpFactory[F] = new NoOpFactory
```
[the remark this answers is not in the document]
seigert: It's not only an import, but also a dependency on `log4cats-noop`, but i see your point. Would constructor method mentioned above be good enough?

We'll be able to use it like
```
abstract class SomeService[F[_]](implicit lf: LggerFactory[F]) { 
 // impl details ... 
}

object SomeServiceTests {
  implicit val lf: LoggerFactory[IO] = NoOpFactory.instance[IO]
  
  // test details ...  
}
```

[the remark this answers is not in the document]
seigert: Removed it, but now I'm not sure about `extends LoggerFactoryGenCompanion`: we'll have it only if user provide it explicitly. And we will happily return any other `LoggerFactoryGen` implicit instance that imports provide...

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
```diff
@@ -0,0 +1,34 @@
...
+
+package org.typelevel.log4cats
+package noop
+
+import cats.Applicative
+
+object NoOpFactory extends LoggerFactoryGenCompanion {
+  def apply[F[_]: Applicative]: LoggerFactory[F] = instanceFor[F]
+
+  def instanceFor[F[_]](implicit F: Applicative[F]): LoggerFactory[F] = new LoggerFactory[F] {
```
danicheg, line 25: To be coherent with `NoOpLogger#impl`.
```suggestion
  def impl[F[_]](implicit F: Applicative[F]): LoggerFactory[F] = new LoggerFactory[F] {
```

danicheg (approved):
Sorry for the delay! Thank you for your work. One minor suggestion from me:

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
seigert: Done!

PR #668, opened 2022-07-02
Upgrade to Scala 3.1, bump base version to 2.3

rossabaker:
A revival of #639.


rossabaker:
#664 does the upgrade

PR #676, opened 2022-08-17
ctxtual

wunderk1nd-e:
- Add helpers for logging with some propagated ctx and kleisli implementation
- Add ce3 module and iolocal instances


build.sbt
```diff
@@ -77,6 +77,16 @@ lazy val noop = crossProject(JSPlatform, JVMPlatform)
...
+lazy val ce3 = crossProject(JSPlatform, JVMPlatform)
+  .settings(commonSettings)
+  .dependsOn(core)
+  .settings(
+    name := "log4cats-ce3",
+    libraryDependencies ++= Seq(
+      "org.typelevel" %%% "cats-effect" % catsEffectV
+    )
+  )
+
```
kubukoz, line 89: doesn't core already depend on CE3?

core/shared/src/main/scala/org/typelevel/log4cats/LoggerFactory.scala
```diff
@@ -48,4 +48,20 @@ object LoggerFactory extends LoggerFactoryGenCompanion {
         fk(logger)
       }
     }
+
```
kubukoz, line 51: I somehow missed that LoggerFactory was a thing - nice!

core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
```diff
@@ -94,6 +96,90 @@ object SelfAwareStructuredLogger {
...
+  def withContextFromKleisli[F[_]: Monad](
+      sl: SelfAwareStructuredLogger[Kleisli[F, Map[String, String], *]]
+  ): SelfAwareStructuredLogger[Kleisli[F, Map[String, String], *]] =
+    withContextF(sl)(
+      Kleisli.ask[F, Map[String, String]]
+    )
+  def withContextF[F[_]: FlatMap](
+      sl: SelfAwareStructuredLogger[F]
+  )(ctx: F[Map[String, String]]): SelfAwareStructuredLogger[F] =
+    new ModifiedContextFSelfAwareStructuredLogger[F](sl)(existingCtx => ctx.map(_ ++ existingCtx))
```
kubukoz, line 108: I think what we're missing is something like this:

```scala
def fa: F[A]

def faWithContext: F[A] = StructuredLogger[F].locally(Map("k" -> "v"))(fa)
```

so that you can wrap an _effect_, not just a logger. That would help propagate the context to whatever services you're calling, without a need to pass the context or the updated logger.

build.sbt
danicheg: Probably we can move this to the `slf4j` module to don't create a new module. Speaking from a user experience it'd be handier - one wouldn't need to add a new dependency. WDYT?

wunderk1nd-e: Happy to go with whatever you guys think is most appropriate - just didn't add it to either of the existing ones because:
- Core doesn't depend on CE3
- Isn't strictly speaking an `slf4j` concern

core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
wunderk1nd-e: How are you imagining something like `StructuredLogger[F].locally` would work?
Just specific implementations for where there's a `IOLocal` or `Kleisli` around?

kubukoz: yeah, I think we'll need concrete implementations

ce3/shared/src/main/scala/org/typelevel/log4cats/ce3/IOLocalHelpers.scala
```diff
@@ -0,0 +1,16 @@
...
+import cats.effect.{IO, IOLocal}
+import org.typelevel.log4cats.{LoggerFactory, SelfAwareStructuredLogger}
+
+object IOLocalHelpers {
+  def loggerWithContextFromIOLocal(
+      sl: SelfAwareStructuredLogger[IO],
+      local: IOLocal[Map[String, String]]
+  ): SelfAwareStructuredLogger[IO] = SelfAwareStructuredLogger.withContextF(sl)(local.get)
+
+  def factoryWithContextFromKleisli(
```
bplommer, line 12: Should this be `FromIOLocal` rather than `FromKleisli`?

wunderk1nd-e: Well spotted 👁
Thanks 🙏

armanbilge:
Instead of adding a CE3 module I think we should just add a CE3 dependency to the core module. CE2 is officially sunset and there's already blocking footguns in core that can only be fixed with CE3.
- https://github.com/typelevel/log4cats/issues/677

build.sbt
lorandszakacs: With #678 we will be depending on cats-effect, and I see no other way to resolve the UUIDGen issue otherwise.

We'll have to track how the dependency on CE in core goes there as well.

armanbilge:
So I had a discussion with @ChristopherDavenport on Discord wrt https://github.com/typelevel/log4cats/issues/677. Although he is supportive of adding a CE dependency to core, he still thinks anything `IOLocal`-based should be in a separate module.

> I would expect that should be in a different module.
> Not core
> Because it's not a way I think folks expect, and the way to do it would be unsafe almost certainly.
> Since they will globalize.
> IOLocal is overspecified. And FiberLocal is a side library.
> So it should be a module or repo with FiberLocal which is a better abstraction then forcing everything to IO

rossabaker:
So the cats-effect-std dependency landed in core, but `IOLocal` would require the full cats-effect.  Doing that would pull cats-effect (and `IO`) into all the downstream projects that limited themselves to cats-effect-std.  That's a pretty good argument for a separate module.

wunderk1nd-e:
> So the cats-effect-std dependency landed in core, but `IOLocal` would require the full cats-effect. Doing that would pull cats-effect (and `IO`) into all the downstream projects that limited themselves to cats-effect-std. That's a pretty good argument for a separate module.

I've updated the branch with `main` - any chance you could approve the workflow run?

build.sbt
rossabaker: Now it depends on `cats-effect-std`.  This still introduces the `IO` dependency.

build.sbt
```diff
@@ -76,6 +76,16 @@ lazy val noop = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   )
   .nativeSettings(commonNativeSettings)
 
+lazy val ce3 = crossProject(JSPlatform, JVMPlatform)
+  .settings(commonSettings)
+  .dependsOn(core)
+  .settings(
+    name := "log4cats-ce3",
```
rossabaker, line 83: I would call this `log4cats-cats-effect`.  Anything beyond `3` will be a breaking change, and I'd avoid colloquial abbrevations in a jar name.

core/shared/src/main/scala/org/typelevel/log4cats/SelfAwareStructuredLogger.scala
rossabaker: I have come to believe that almost all uses of `IOLocal` should have that `cats.mtl.Local` shape.  See https://github.com/typelevel/cats-effect/issues/3385.  And I like the `scope` name: that's what `Local` calls it.

rossabaker (commented):
I apologize for losing track of this one... I approved the CI run and resolved a merge conflict.

armanbilge:
> > I have come to believe that almost all uses of `IOLocal` should have that `cats.mtl.Local` shape. See [typelevel/cats-effect#3385](https://github.com/typelevel/cats-effect/issues/3385). And I like the `scope` name: that's what `Local` calls it.

Should this PR then be in terms of `Local`, instead of offering specific instances?

rossabaker:
I would have thought so, until cats-mtl's future stability was questioned on [this pull request](https://github.com/typelevel/cats-effect/pull/3429).  The uncertainty around that statement is blocking a cats-effect PR, a clear path forward on otel4s, and now this.

armanbilge:
Yeah. I know it's annoying, but might be worth pausing on this one as well until that's decided. We already have a CE dependency in log4cats core, and CE is already (technically) tied to MTL versioning, and thus so are we.

rossabaker:
If @wunderk1nd-e isn't desperate for a release on that, I think that's a sound strategy.  The fact everything is being done twice is because there's a lovely type class for this if we'd accept it as a routine dependency.

mattyjbrown:
Is it worth it/feasible to add a plain and simple `LoggerFactory.withContext`, as a way of propagating context when you're passing around `LoggerFactory`s, or is that going to conflict the eventual implementation of this PR?

armanbilge:
CE v3.6.0 was finally released with a dependency on Cats MTL and a `Local` instance for `IOLocal`. So we can finish up this PR.

Jay-Lokhande:
Hey, I'll be looking into this PR and trying to make the final changes. I'll update here as I progress.

morgen-peschke:
Closing in favor of it's continuation in #905

Issue #677, opened 2022-08-20
`randomUUID` is blocking


armanbilge:
https://github.com/typelevel/log4cats/blob/4f02ef39fbee3568c9eaa52fee42904247988c62/core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala#L104

See extensive discussion in https://github.com/typelevel/cats-effect/issues/2865#issuecomment-1061950448.

The best thing to do here is use `cats.effect.std.UUIDGen`.

PR #678, opened 2022-08-20
Use `UUIDGen` instead of `randomUUID()`

armanbilge:
Fixes https://github.com/typelevel/log4cats/issues/677.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ val Scala213 = "2.13.8"
 val Scala212 = "2.12.16"
 val Scala3 = "3.1.3"
 
-ThisBuild / tlBaseVersion := "2.4"
+ThisBuild / tlBaseVersion := "2.5"
 ThisBuild / crossScalaVersions := Seq(Scala213, Scala212, Scala3)
 ThisBuild / scalaVersion := Scala213
 ThisBuild / startYear := Some(2018)
--- a/build.sbt
+++ b/build.sbt
@@ -46,7 +46,7 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
   .settings(
     name := "log4cats-core",
     libraryDependencies ++= Seq(
-      "org.typelevel" %%% "cats-core" % catsV,
+      "org.typelevel" %%% "cats-core"       % catsV,
       "org.typelevel" %%% "cats-effect-std" % catsEffectV
     ),
     libraryDependencies ++= {
--- a/build.sbt
+++ b/build.sbt
@@ -46,17 +46,13 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
   .settings(
     name := "log4cats-core",
     libraryDependencies ++= Seq(
-      "org.typelevel" %%% "cats-core" % catsV
+      "org.typelevel" %%% "cats-core" % catsV,
+      "org.typelevel" %%% "cats-effect-std" % catsEffectV
     ),
     libraryDependencies ++= {
       if (tlIsScala3.value) Seq.empty
       else Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided)
     }
-  )
-  .jsSettings(
-    // https://www.scala-js.org/news/2022/04/04/announcing-scalajs-1.10.0#fixes-with-compatibility-concerns
-    libraryDependencies += ("org.scala-js" %%% "scalajs-java-securerandom" % "1.0.0")
-      .cross(CrossVersion.for3Use2_13)
   )
 
 lazy val testing = crossProject(JSPlatform, JVMPlatform)
--- a/core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
@@ -17,7 +17,8 @@
 package org.typelevel.log4cats
 
 import cats._
-import cats.implicits._
+import cats.effect.std.UUIDGen
+import cats.syntax.all._
 
 import java.io.{PrintWriter, StringWriter}
 import java.util.UUID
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -46,18 +46,14 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
...
     libraryDependencies ++= {
       if (tlIsScala3.value) Seq.empty
       else Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value % Provided)
     }
   )
-  .jsSettings(
-    // https://www.scala-js.org/news/2022/04/04/announcing-scalajs-1.10.0#fixes-with-compatibility-concerns
-    libraryDependencies += ("org.scala-js" %%% "scalajs-java-securerandom" % "1.0.0")
-      .cross(CrossVersion.for3Use2_13)
-  )
```
armanbilge, line 60: Although in theory this dependency is safe, in practice it is not safe to rely on. See discussion in:
- https://github.com/typelevel/cats-effect/pull/2945

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -52,12 +53,26 @@ object PagingSelfAwareStructuredLogger {
    * @return
    *   SelfAwareStructuredLogger with pagination.
    */
-  def withPaging[F[_]: Monad](pageSizeK: Int = 64, maxPageNeeded: Int = 999)(
+  def withPaging2[F[_]: Monad: UUIDGen](pageSizeK: Int = 64, maxPageNeeded: Int = 999)(
```
armanbilge, line 56: Sorry, I wasn't feeling creative with naming at this moment 😅 I will happily change this to any suggestion, thanks!

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
```diff
@@ -52,12 +53,26 @@ object PagingSelfAwareStructuredLogger {
...
 
-  private class PagingSelfAwareStructuredLogger[F[_]: Monad](pageSizeK: Int, maxPageNeeded: Int)(
+  @deprecated("Use withPaging2", "2.5.0")
+  def withPaging[F[_]: Monad](pageSizeK: Int = 64, maxPageNeeded: Int = 999)(
+      logger: SelfAwareStructuredLogger[F]
+  ): SelfAwareStructuredLogger[F] =
+    new PagingSelfAwareStructuredLogger[F](
+      pageSizeK,
+      maxPageNeeded,
+      Monad[F].unit.map(_ => UUID.randomUUID())
```
armanbilge, line 68: Terrible, right! But this just more clearly exposes the bad behavior that already existed here. There is no safe way to generate random `UUID`s with only a `Monad`.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
lorandszakacs: Maybe we should name it `paged`? 🤔 and just keep that as the future API

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
lorandszakacs: Absolutely! I'm just sorry I didn't catch it before

lorandszakacs:
LGTM! I'm fine with merging this. Leaving it open for other maintainers to weigh in.

core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala
armanbilge: Works for me! :)

lorandszakacs (approved):

build.sbt
```diff
@@ -46,18 +46,14 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
   .settings(
     name := "log4cats-core",
     libraryDependencies ++= Seq(
-      "org.typelevel" %%% "cats-core" % catsV
+      "org.typelevel" %%% "cats-core"       % catsV,
+      "org.typelevel" %%% "cats-effect-std" % catsEffectV
```
danicheg, line 50: Just to keep the initial goal of having the core module CE-less, can we introduce an internal interface for UUIDGen and its implementation within the slf4j module?

lorandszakacs: Can we instead create a new module that is _explicitely_ dependent on CE? And move the correct version of this there?

I feel like re-defining a UUIDGen equivalent leads to a lot of confusion and bad UX. Instead having a `log4cats-core` package that is honest about its limitations, and an extension `logs4cats-ce` (naming is hard) that overcomes aforementioned limitations might be better.

Although not entirely sure how to pull this off in a bincompat _and_ user friendly way.

armanbilge: I agree, adding a "`log4cats-effect`" would be better for UX, but still not great for UX+bincompat.

Just so that we are clear: what are the arguments against just have a CE dependency in the core module?

lorandszakacs: > what are the arguments against just have a CE dependency in the core module?

Very good question! I personally don't know since I probably missed those discussions during my absence. :see_no_evil: Personally I think it's ok to have CE in `core`

danicheg: I can only guess, but perhaps the reason is to apply the Rule of Least Power. But once again, I'm not sure if this is a thing still.

armanbilge: I would say it is a thing still :) The problem is the repo is currently applying the principle of too-little-power. This is a course-correction.

armanbilge:
@danicheg so wdyt, can we move forward with adding CE to core?

If not, my best suggestion is to:
1. Create a "`log4cats-effect`" module.
2. Completely deprecate `log4cats.PagingSelfAwareStructuredLogger` as blocking and unfixable.
3. Re-introduce it as `log4cats.effect.PagingSelfAwareStructuredLogger` or similar.

danicheg:
To sum up my thoughts:
I haven't a strict 'no' about having CE in the core module. 
I was wondering – is it possible to preserve the initial goal of not having CE in the core module? Perhaps the only reason why this could be still a thing – is the principle of least power. That said if we will benefit from adding this dependency, then probably, we are good. 
It'd be extremely helpful if initial architectures of log4cats would have a glance at this one. But based on common sense, I don't see much of the evil in adding CE to the core module.

armanbilge:
I spoke with @ChristopherDavenport on Discord. Here's a quick summary:

Regarding CE in core:
> Having cats-effect to in core I don't mind though. The idea was a bunch of abstractions and helpers there, with the effects of them in the helper modules. But... If we need the std to have cats-effect for those helpers. I'm fine with that.
> Basically, core is like http4s core. (I reuse concepts)

Regarding UUIDs:
> It's to identify all the split logs as a single log action.
> So it's really just a string.
> So I'm not seeing why it wouldn't be `F[String]` and let people define their own way, one of which could be via UUID and UuidGen

---

So it boils down to a question of UX:
1. We can keep core on Cats only, and force users to bring their own `F[String]` since we cannot provide a default.
2. We can move ahead with this PR, which adds Cats Effect, and provides a default `F[String]` via `GenUUID[F]`. Then in a follow-up we can make this customizable (or I can do it in this PR if it's easy).

Thoughts?

@danicheg thanks again for raising this issue, it's opened some good questions.

danicheg:
It comes that all of us don't mind adding the CE std module to the log4cats core. So let's move onward with this. :+1: from me to that.

Issue #680, opened 2022-08-22
Implement a Scala.js `ConsoleLogger`


armanbilge:
There's currently no "default" logging backend for Scala.js, but it should be easy to add one based on the JavaScript `console` APIs. Technically `console` is not part of core JavaScript, but it is specced and is fairly standard across Node.js and browsers.

- https://console.spec.whatwg.org/
- https://developer.mozilla.org/en-US/docs/Web/API/console
- https://nodejs.org/api/console.html

I'm not a maintainer on this repo so if someone works on this or has questions please make sure to ping me :)

armanbilge:
Here's a partial facade for the `console` APIs. We should inline it instead of adding a dependency to scala-js-dom because that dependency makes sense only for browsers.

https://github.com/scala-js/scala-js-dom/blob/51a807cc04ea134ece030d6d4005a8ce30c1de99/dom/src/main/scala/org/scalajs/dom/Console.scala

2chilled:
@danicheg It seems like there is no release including this functionality available yet. Can we change that? Would be great!

armanbilge:
Yes, I'd love a release with this too. Until then, there is a snapshot.
https://s01.oss.sonatype.org/content/repositories/snapshots/org/typelevel/log4cats-js-console_sjs1_3/2.5.0-99-e87a0cd-SNAPSHOT/

rossabaker:
I just created a 3.6.0 milestone with the things that are almost ready, and we can get a new release.

PR #681, opened 2022-08-22
Add `NoOpFactory` implicit in package object

armanbilge:
This is the no-op equivalent of:
https://github.com/typelevel/log4cats/blob/4f02ef39fbee3568c9eaa52fee42904247988c62/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/package.scala#L23


danicheg (commented):
We intentionally didn't add an implicit method when adding `NoOpFactory`, [see](https://github.com/typelevel/log4cats/pull/667#pullrequestreview-1023423720). I personally beware of this kind of thing since it accidentally may hurt someone in $PROD. Especially considering that some $IDE can suggest import when implicit value is required. But I can be overthinking about this. So, if the majority will be happy with this implicit method, I will be happy too.

armanbilge:
> Especially considering that some $IDE can suggest import when implicit value is required.

If this is so dangerous then I argue that `LoggerFactory` actually should not be an implicit at all. Since there can be multiple implementations, doesn't this go against our philosophy anyway?

danicheg:
> Since there can be multiple implementations, doesn't this go against our philosophy anyway?

Sorry, I don't quite follow your idea. I mean that if this method is just for tests, then it shouldn't be too hard to create an implicit instance on the user side. Or, probably, we can add this method straightforwardly to the `testing` module. Then it'd be a little safer from accidental bugs. But again, all of this could be just my imagination.

armanbilge:
Ok, fair enough. My motivation is that currently there is no `LoggerFactory` for Scala.js. But maybe better to solve that problem directly.
- https://github.com/typelevel/log4cats/issues/680

We will have a bigger problem for Scala Native, because there is no standard logging framework there.

armanbilge:
> Or, probably, we can add this method straightforwardly to the `testing` module.

To be fair, it's already isolated to the no-op module. If no-op was part of core that would be a bigger problem, but seems just as hard to add a no-op dependency as it is to add testing dependency :)

danicheg:
> If no-op was part of core that would be a bigger problem, but seems just as hard to add a no-op dependency as it is to add testing dependency :)

Yeah, true. Probably initial authors considering this fine. So I'm speaking just as a random guy from the community 🙇🏻‍♂️

lorandszakacs:
LGTM. The implicit `LoggerFactory` for a `Sync` in `slf4j ` module is way more dangerous :D and if we live with that we can at least be consistent about it.

lorandszakacs (approved):

lorandszakacs:
I also lean towards deprecating the implicit creation of the factory in slf4j, in which case this indeed should be closed 😄

But I do not agree with encouraging people to not pass `LoggerFactory` implicitly, only with its automagic implicit creation. At the end of the day how many users will have more than one instance of a LoggerFactory in their codebase? And if they do, there always various ways of newtype-ing, or subclassing it, etc. to solve the problem. For large codebases with literally hundreds of classes 90% of which do logging users will probably pass these around implicitly.

bplommer:
Why not provide this as a convenience but in an `implicits` namespace, like we do with (for example) the cats-effect global runtime? That way it's not just "floating around out there" but it's available in a convenient form when needed.

lorandszakacs:
@armanbilge Shall we close this PR, and see if @bplommer's idea lands either in #683 or a follow-up?

armanbilge:
Ah yes, I think we're all agreed we don't want to do this.

PR #683, opened 2022-08-23
Deprecate implicit materialization of `Slf4jFactory`

armanbilge:
See https://github.com/typelevel/log4cats/pull/681#issuecomment-1224535985. Since there is not one canonical `LoggerFactory` implementation, they should be passed explicitly, not implicitly.

Closes https://github.com/typelevel/log4cats/pull/681.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -100,7 +100,7 @@ This brings several advantages:
 If you are unsure how to create a new `LoggerFactory[F]` instance, then you can look at the `log4cats-slf4j`,
 or `log4cats-noop` modules for concrete implementations.
 
-The quickest fix might to create an instance of a `Slf4jFactory` and pass it around implicitly:
+The quickest fix might be to create an instance of a `Slf4jFactory` and pass it around implicitly:
 ```scala mdoc:reset:silent
 import cats.effect.IO
 import cats.Monad
--- a/docs/index.md
+++ b/docs/index.md
@@ -100,29 +100,17 @@ This brings several advantages:
 If you are unsure how to create a new `LoggerFactory[F]` instance, then you can look at the `log4cats-slf4j`,
 or `log4cats-noop` modules for concrete implementations.
 
-The quickest fix might be to import needed implicits:
-```scala mdoc:silent
-// assumes dependency on log4cats-slf4j module
-import org.typelevel.log4cats._
-import org.typelevel.log4cats.slf4j._
-
-val logger: SelfAwareStructuredLogger[IO] = LoggerFactory[IO].getLogger
-
-// or
-def anyFSyncLogger[F[_]: Sync]: SelfAwareStructuredLogger[F] = Slf4jFactory[F].getLogger
-```
-
-Alternatively, a mutually exclusive solution is to explicitly create your 
-`LoggerFactory[F]` instance and pass them around implicitly:
+The quickest fix might to create an instance of a `Slf4jFactory` and pass it around implicitly:
 ```scala mdoc:reset:silent
 import cats.effect.IO
 import cats.Monad
 import cats.syntax.all._
 import org.typelevel.log4cats._
+// assumes dependency on log4cats-slf4j module
 import org.typelevel.log4cats.slf4j.Slf4jFactory
 
 // create our LoggerFactory
-implicit val logging: LoggerFactory[IO] = Slf4jFactory[IO]
+implicit val logging: LoggerFactory[IO] = Slf4jFactory.create[IO]
 
 // we summon LoggerFactory instance, and create logger
 val logger: SelfAwareStructuredLogger[IO] = LoggerFactory[IO].getLogger
--- a/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jFactory.scala
+++ b/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jFactory.scala
@@ -17,6 +17,7 @@
 package org.typelevel.log4cats
 package slf4j
 
+import cats.effect.kernel.Sync
 import org.slf4j.{Logger => JLogger}
 
 trait Slf4jFactory[F[_]] extends LoggerFactory[F] {
@@ -27,6 +28,20 @@ trait Slf4jFactory[F[_]] extends LoggerFactory[F] {
 
 object Slf4jFactory extends LoggerFactoryGenCompanion {
   def apply[F[_]: Slf4jFactory]: Slf4jFactory[F] = implicitly
+
+  def create[F[_]: Sync]: Slf4jFactory[F] = new Slf4jFactory[F] {
+    override def getLoggerFromName(name: String): SelfAwareStructuredLogger[F] =
+      Slf4jLogger.getLoggerFromName(name)
+
+    override def getLoggerFromSlf4j(logger: JLogger): SelfAwareStructuredLogger[F] =
+      Slf4jLogger.getLoggerFromSlf4j(logger)
+
+    override def fromName(name: String): F[SelfAwareStructuredLogger[F]] =
+      Slf4jLogger.fromName(name)
+
+    override def fromSlf4j(logger: JLogger): F[SelfAwareStructuredLogger[F]] =
+      Slf4jLogger.fromSlf4j(logger)
+  }
 
   def getLoggerFromSlf4j[F[_]](logger: JLogger)(implicit
       lf: Slf4jFactory[F]
--- a/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jFactory.scala
+++ b/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jFactory.scala
@@ -17,6 +17,7 @@
 package org.typelevel.log4cats
 package slf4j
 
+import cats.effect.kernel.Sync
 import org.slf4j.{Logger => JLogger}
 
 trait Slf4jFactory[F[_]] extends LoggerFactory[F] {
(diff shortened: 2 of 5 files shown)


slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jFactory.scala
```diff
@@ -26,8 +27,23 @@ trait Slf4jFactory[F[_]] extends LoggerFactory[F] {
 }
 
 object Slf4jFactory extends LoggerFactoryGenCompanion {
+  @deprecated("LoggerFactories should be passed explicitly, not implicitly", "2.5.0")
   def apply[F[_]: Slf4jFactory]: Slf4jFactory[F] = implicitly
 
+  def create[F[_]: Sync]: Slf4jFactory[F] = new Slf4jFactory[F] {
```
armanbilge, line 33: This would have been more convenient as an `apply` method, but unfortunately that's been squatted :(

lorandszakacs:
To me `LoggerFactory` is akin to capability traits like `cats.effect.std.Console` and `fs2.io.Network`, and _can_ be passed around implicitely as here in [skunk](https://github.com/tpolecat/skunk/blob/main/modules/core/shared/src/main/scala/Session.scala#L296).

It's just that I would highly encourage _explicit_ instantiation of these by the user, and then allow them to chose if they wish to pass implicitely or explicitely in their own app :). I think the status quo is that `Console` is fine to be passed implicitely, right? If so, I'd reword the deprecation message on all methods to indicate that explicit instantation of a LoggerFactory of the user's choosing + the calling of methods directly on said LoggerFactory.

I'd also prefer to see this paired with some docs that showcase the explicit  use of  `Slf4jFactory.create` and passing it along to some classes (both implicitely and explicitely). Basically something like:

```scala
// the assorted imports, also code does not compile
object Main extends IOApp.Simple {
  override def run: IO[Unit] = for {
     implicit0(lf: LoggerFactory[IO]) = Slf4jFactory.create[IO]
     // other capabilities like cats.effect.std.{Console, Random}, etc here.
     app <- instantiateApp
  } yield ()
  
  type MyApp = Nothing

  def instantiateApp[F[_]: LoggerFactory]: IO[MyApp] = 
   for {
      // many others in IO
      someClass = new SomeClass()
      // many others in IO
   } yield ???
}

class SomeClass[F[_]](implicit loggers: LoggerFactory[F]){
   private val logger = loggers.getLogger //explicit call the LoggerFactory passed along by the user, even if implicit
}
```

lorandszakacs:
Maybe we should make some kind of integrative typelevel doc that showcases _a_ pattern of instantiating capabilities available in the ecosystem and link _that_ as an example? 😅 

Since it would serve well to be consistent across everything under `org.typelevel` and people would also have a handle on _why_ we do something like this in log4cats, fs2, cats-effect, etc.

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jFactory.scala
lorandszakacs: I think it's good this way! `cats.effect.std.Console` is also instatiated using `make` instead of `apply`. Maybe we should rename this to `make` rather than `create` :D

armanbilge:
> To me `LoggerFactory` is akin to capability traits like `cats.effect.std.Console` and `fs2.io.Network`, and _can_ be passed around implicitely as here in [skunk](https://github.com/tpolecat/skunk/blob/main/modules/core/shared/src/main/scala/Session.scala#L296).

I definitely agree those examples can be passed around implicitly :) that's because they have _canonical_ implementations. Meaning there there is essentially only one true implementation.

- are you aware of any library providing an implementation of `Console` besides [the one in Cats Effect](https://github.com/typelevel/cats-effect/blob/3da61a59438da30f9bb01192fec9e690aacc3ce4/std/shared/src/main/scala/cats/effect/std/ConsoleCrossPlatform.scala#L176)? What would such an implementation do?
- `Network` is in fact [`sealed`](https://github.com/typelevel/fs2/blob/0e573f2ca899886af57aaea9edd7a1072b4eb00a/io/shared/src/main/scala/fs2/io/net/Network.scala#L49) and therefore fs2 can provide the one and only canonical implementation

This is not the case for `LoggerFactory`, and indeed would be quite inconvenient if it was :)

The fact that there are legitimate fears in https://github.com/typelevel/log4cats/pull/681#pullrequestreview-1081706224 that the wrong/unexpected implementation could be passed implicitly reinforces that this should be explicit.

---

I definitely agree the docs can/should be improved (indeed CI is failing about that). But first let's make sure we're all on the same page with this change :)

armanbilge:
Ah, your comment in https://github.com/typelevel/log4cats/pull/681#issuecomment-1224676164 clarifies things for me.

> But I do not agree with encouraging people to not pass `LoggerFactory` implicitly, only with its automagic implicit creation.

That seems reasonable to me: we'd deprecate just this method and nothing more.
https://github.com/typelevel/log4cats/blob/d00181c9e350dd538b86676cd3f6f209228abfcd/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/package.scala#L24

Happy to hear other's thoughts :)

lorandszakacs:
> The fact that there are legitimate fears in https://github.com/typelevel/log4cats/pull/681#pullrequestreview-1081706224 that the wrong/unexpected implementation could be passed implicitly reinforces that this should be explicit.

Yes, but the fear is accidentally getting a new `LoggerFactory` created implicitely, not that it _is_ passed around implicitely. I detailed the reasoning on [this comment](https://github.com/typelevel/log4cats/pull/681#issuecomment-1224676164) in #681. I'll copy paste the relevant bits.

> At the end of the day how many users will have more than one instance of a LoggerFactory in their codebase? And if they do, there always various ways of newtype-ing, or subclassing it, etc. to solve the problem. For large codebases with literally hundreds of classes 90% of which do logging users will probably pass these around implicitly.

> that's because they have canonical implementations

Well, there are _good_ implementations of the capabilities. And I definitely would not be the one to try to implement a `fs2.io.Network` differently 😅. But they are hardly _canonical_ in the way implementing `Monad` for `Option` is.

lorandszakacs:
> That seems reasonable to me: we'd deprecate just this method and nothing more.

Then we are in agreement 🥳

lorandszakacs (approved):

docs/index.md
```diff
@@ -100,29 +100,17 @@ This brings several advantages:
...
-
-val logger: SelfAwareStructuredLogger[IO] = LoggerFactory[IO].getLogger
-
-// or
-def anyFSyncLogger[F[_]: Sync]: SelfAwareStructuredLogger[F] = Slf4jFactory[F].getLogger
-```
-
-Alternatively, a mutually exclusive solution is to explicitly create your 
-`LoggerFactory[F]` instance and pass them around implicitly:
+The quickest fix might to create an instance of a `Slf4jFactory` and pass it around implicitly:
```
armanbilge, line 103:
```suggestion
The quickest fix might be to create an instance of a `Slf4jFactory` and pass it around implicitly:
```

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jFactory.scala
armanbilge: Yes, agree now that we decided we are not totally deprecating `implicit` (otherwise, the `apply` was dead deprecated weight).

Regarding naming: I chose `create` based on this.
https://github.com/typelevel/log4cats/blob/4f02ef39fbee3568c9eaa52fee42904247988c62/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala#L38

But I think `make` is good too 🤔  and maybe better to keep it independent of `create`.

lorandszakacs (approved):

bplommer:
To repeat my comment in https://github.com/typelevel/log4cats/pull/681#issuecomment-1231434973 - Why not provide this as a convenience but in an implicits namespace, like we do with (for example) the cats-effect global runtime? That way it's not just "floating around out there" but it's available in a convenient form when needed.

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/package.scala
```diff
@@ -17,20 +17,10 @@
...
-    override def getLoggerFromSlf4j(logger: JLogger): SelfAwareStructuredLogger[F] =
-      Slf4jLogger.getLoggerFromSlf4j(logger)
-
-    override def fromName(name: String): F[SelfAwareStructuredLogger[F]] =
-      Slf4jLogger.fromName(name)
-
-    override def fromSlf4j(logger: JLogger): F[SelfAwareStructuredLogger[F]] =
-      Slf4jLogger.fromSlf4j(logger)
-  }
+  @deprecated("Use Slf4jFactory.create[F] explicitly", "2.5.0")
```
lorandszakacs, line 23: Version needs to be bumped to 2.6.0 now 😅

armanbilge, line 23:
```suggestion
  @deprecated("Use Slf4jFactory.create[F] explicitly", "2.6.0")
```

lorandszakacs:
> To repeat my comment in [#681 (comment)](https://github.com/typelevel/log4cats/pull/681#issuecomment-1231434973) - Why not provide this as a convenience but in an implicits namespace, like we do with (for example) the cats-effect global runtime? That way it's not just "floating around out there" but it's available in a convenient form when needed.

@danicheg what do you think about this approach?

I'm fine going forward with this even without convenience `.implicits` import. But I'm not against it.

danicheg (approved):

danicheg:
> > To repeat my comment in https://github.com/typelevel/log4cats/pull/681#issuecomment-1231434973 - Why not provide this as a convenience but in an implicits namespace, like we do with (for example) the cats-effect global runtime? That way it's not just "floating around out there" but it's available in a convenient form when needed.
> 
> @danicheg what do you think about this approach?

I think that `Runtime` abstraction and providing instances for `LoggerFactory` through the implicit scope – differ in their nature. But I'm biased in this question. I don't like the approach of using logging via capabilities.

lorandszakacs:
Merging and we'll defer the discussion about implicit LoggerFactory when the ecosystem reaches more of a consensus about implicit capabilities in general :sweat_smile:

PR #693, opened 2022-09-28
Beginner Friendly Example in README

mjohns39:
Added a more beginner friendly example and explanation to the README to help them bridge the gap from vanilla logging to log4cats

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,23 @@ libraryDependencies ++= Seq(
 )
 ```
 
+## Why log4cats?
+
+Well, to answer that, let's take a look at how you might combine cats-effect with vanilla logging...
+
+```scala
+object MyVanillaLoggingThing {
+  val logger: Logger = Logger(LoggerFactory.getLogger(getClass.getName))
+  
+  def doSomething(): IO[Unit] =
+    IO(logger.info("Doing something!") *> IO.println("Hello, World!")
+
+}
+```
+
+But what if you don't want to wrap your logger in an `IO` like this?  
+Good news, you don't have to!  Enter log4cats!  Read on!
+
 ## Examples
 
 ```scala
--- a/README.md
+++ b/README.md
@@ -73,6 +73,20 @@ object MyThing {
 }
 ```
 
+## Wait...why log4cats again?
+If you compare the vanilla logger + cats-effect example with the log4cats examples above,
+you might be asking yourself "why do I have to do any of this?" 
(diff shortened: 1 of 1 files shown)


mjohns39:
Opening this following a discussion in TypeLevel Discord server:  https://discord.com/channels/632277896739946517/632278585700384799/1024710118023114824

README.md
```diff
@@ -18,6 +18,24 @@ libraryDependencies ++= Seq(
 )
 ```
 
+## Why log4cats?
+
+Short answer is you don't really need log4cats if you already have logging setup in your app.
```
danicheg, line 23: > – Why log4cats?
> – Short answer is you don't really need log4cats...

I'm sorry, but that totally creased me 😆 
Can we change this to something less disappointing for maintainers? 🙇🏻‍♂️

mjohns39: Done.  Let me know what you think of the new version :)

mjohns39: P.S.  I squashed the commits

README.md
```diff
@@ -18,6 +18,23 @@ libraryDependencies ++= Seq(
 )
 ```
 
+## Why log4cats?
+
```
lorandszakacs, line 22: I'd also include another sentence of the sort:
```
Logging is oftimes an overlooked side-effect, which under the hood can range from writing to a mutable queue, writing to disk, outputing to the console, or sometimes even doing network I/O! To correctly deal with these kinds of side-effects we have to ensure they are properly wrapped in `IO`, see the [cats-effect docs](https://typelevel.org/cats-effect/docs/concepts#side-effects) for more details.
```

And then tie it in with the sentence you wrote.

Now I realize this essentially doubles the amounts of words :sweat_smile: it definitely requires _some_ editing :D, and I'm not clinging to any particular form or formulation. Just want to point readers to the fact that there is documentation around dealing with side-effects.

mjohns39: I see your point.  The purpose of this example was targeted towards new learners and people possibly in a hybrid state that want to be using cats-effect with their existing app.  The goal was to give something easy to understand, but also valid with the cats-effect ethos.  So I don't think I should include it in the "Why log4cats?" section...but I will add it further down

mjohns39: Pushed change.  Let me know what you think?

lorandszakacs: Nice!

lorandszakacs (approved):
Double checked for typos.

Really like the `## Wait...why log4cats again?` section :seedling:

Issue #718, opened 2023-01-09
Provide Writer/WriterT backed StructuredLogger


morgen-peschke:
The idea behind `WriterLogger` and `WriterTLogger` is really good. Unfortunately, they're not directly usable when taking advantage of capability-based logging as that requires a `StructuredLogger` and they are `SelfAwareLogger`s.

morgen-peschke:
Closing issue as PR has been merged.

PR #719, opened 2023-01-09
Add structured equivalents of Writer & WriterT loggers

morgen-peschke:
Adds `WriterStructuredLogger`  and `WriterTStructuredLogger` as equivalents of `WriterLogger` and `WriterTLogger`, respectively.

Implements #718

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterLogger.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterLogger.scala
@@ -35,89 +35,14 @@ object WriterLogger {
       infoEnabled: Boolean = true,
       warnEnabled: Boolean = true,
       errorEnabled: Boolean = true
-  ): SelfAwareLogger[Writer[G[LogMessage], *]] = {
-    implicit val monoidGLogMessage: Monoid[G[LogMessage]] = Alternative[G].algebra[LogMessage]
-    new SelfAwareLogger[Writer[G[LogMessage], *]] {
-      def isTraceEnabled: Writer[G[LogMessage], Boolean] =
-        Writer.value[G[LogMessage], Boolean](traceEnabled)
-      def isDebugEnabled: Writer[G[LogMessage], Boolean] =
-        Writer.value[G[LogMessage], Boolean](debugEnabled)
-      def isInfoEnabled: Writer[G[LogMessage], Boolean] =
-        Writer.value[G[LogMessage], Boolean](infoEnabled)
-      def isWarnEnabled: Writer[G[LogMessage], Boolean] =
-        Writer.value[G[LogMessage], Boolean](warnEnabled)
-      def isErrorEnabled: Writer[G[LogMessage], Boolean] =
-        Writer.value[G[LogMessage], Boolean](errorEnabled)
-
-      def debug(t: Throwable)(message: => String): Writer[G[LogMessage], Unit] =
-        if (debugEnabled)
-          Writer.tell(Alternative[G].pure(LogMessage(LogLevel.Debug, t.some, message)))
-        else Writer.value[G[LogMessage], Unit](())
-      def error(t: Throwable)(message: => String): Writer[G[LogMessage], Unit] =
-        if (errorEnabled)
-          Writer.tell(Alternative[G].pure(LogMessage(LogLevel.Error, t.some, message)))
-        else Writer.value[G[LogMessage], Unit](())
-      def info(t: Throwable)(message: => String): Writer[G[LogMessage], Unit] =
-        if (infoEnabled)
-          Writer.tell(Alternative[G].pure(LogMessage(LogLevel.Info, t.some, message)))
-        else Writer.value[G[LogMessage], Unit](())
-      def trace(t: Throwable)(message: => String): Writer[G[LogMessage], Unit] =
-        if (traceEnabled)
-          Writer.tell(Alternative[G].pure(LogMessage(LogLevel.Trace, t.some, message)))
-        else Writer.value[G[LogMessage], Unit](())
-      def warn(t: Throwable)(message: => String): Writer[G[LogMessage], Unit] =
-        if (warnEnabled)
-          Writer.tell(Alternative[G].pure(LogMessage(LogLevel.Warn, t.some, message)))
-        else Writer.value[G[LogMessage], Unit](())
-      def debug(message: => String): Writer[G[LogMessage], Unit] =
-        if (debugEnabled)
-          Writer.tell(Alternative[G].pure(LogMessage(LogLevel.Debug, None, message)))
-        else Writer.value[G[LogMessage], Unit](())
-      def error(message: => String): Writer[G[LogMessage], Unit] =
-        if (errorEnabled)
-          Writer.tell(Alternative[G].pure(LogMessage(LogLevel.Error, None, message)))
-        else Writer.value[G[LogMessage], Unit](())
-      def info(message: => String): Writer[G[LogMessage], Unit] =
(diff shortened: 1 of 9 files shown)


core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterTStructuredLogger.scala
```diff
@@ -0,0 +1,153 @@
...
+/**
+ * >>> WARNING READ BEFORE USAGE! <<< This logger will NOT log anything if `F` fails!
+ *
+ * Running the `WriterT` instance will yield a value of type `F[(G[LogMessage], A)]`. As a result,
+ * the logged messages can be materialized if and only `F` succeeds.
+ *
+ * This is particularly important because of the way that `cats.effect.IO` handles cancellation and
+ * timeouts.
+ *
+ * If a `SelfAwareStructuredLogger` is needed for testing, the `testing` module provides a better
```
morgen-peschke, line 34: @ChristopherDavenport how does this redirect look?

testing/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+class ExampleTest extends CatsEffectSuite {
+  private val loggerFixture = ResourceFunFixture[LoggingHelper]((options: TestOptions) =>
+    Resource.make(IO.delay {
+      val logger = StructuredTestingLogger.impl[IO]()
+      LoggingHelper(logger, logger.addContext(Map("TestName" -> options.name)))
+    })(_ => IO.unit)
+  )
+
+  private val loggerPrinter = new Printer {
```
morgen-peschke, line 41: There's rather a lot of boilerplate in this. It might be worth considering creating an integration for `munit`.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterStructuredLogger.scala
```diff
@@ -0,0 +1,55 @@
...
+      debugEnabled,
+      infoEnabled,
+      warnEnabled,
+      errorEnabled
+    )
+
+  def run[F[_]: Applicative, G[_]: Foldable](
+      l: StructuredLogger[F]
+  ): Writer[G[StructuredLogMessage], *] ~> F =
+    new ~>[Writer[G[StructuredLogMessage], *], F] {
```
rossabaker, line 49: Absolutely not a big deal, but I usually write these with infix notation, just like the types.

```scala
new (Writer[G[StructuredLogMessage], *] ~> F) {
```

testing/README.md
rossabaker: We're on a milestone, for reasons I can't remember.  If that becomes a production dependency, we'd probably want a separate repo.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterTStructuredLogger.scala
```diff
@@ -0,0 +1,153 @@
...
+import org.typelevel.log4cats.{SelfAwareStructuredLogger, StructuredLogger}
+
+/**
+ * >>> WARNING READ BEFORE USAGE! <<< This logger will NOT log anything if `F` fails!
+ *
+ * Running the `WriterT` instance will yield a value of type `F[(G[LogMessage], A)]`. As a result,
+ * the logged messages can be materialized if and only `F` succeeds.
+ *
+ * This is particularly important because of the way that `cats.effect.IO` handles cancellation and
+ * timeouts.
```
rossabaker, line 32: Good reasons not to use it, but no reason to use it.  Maybe highlight that it works without a Sync instance?

rossabaker (commented):
I like it, for surfacing the test code as much as the WriterT.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterTStructuredLogger.scala
rossabaker: Better way to look at it: if your log values are not a global operational log, but annotations with referential integrity to an entity, this is may be a good choice.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterStructuredLogger.scala
morgen-peschke: Fair 👍🏻 
This is nearly a direct copy of the existing one, so if you're of the mind to clean this up at some point, it's probably worth changing those to keep things consistent.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterTStructuredLogger.scala
morgen-peschke: I'll have a think on how to word that

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterStructuredLogger.scala
rossabaker: Yeah, let's just keep these consistent with what we already have.  My way may be idiosyncratic, but it's consistent with how we usually express the return type.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterTStructuredLogger.scala
rossabaker: Yeah, mine still isn't great.  The key difference, for better and for worse, is that it matches the life cycle of the value.  If the logs are relevant in the absence of the value, this is very bad.  If the logs are tightly coupled to the value and need not be retained longer than the value, this is highly convenient!

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterStructuredLogger.scala
morgen-peschke: I ended up updating this when I went in to update the positive side of the warnings.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterTStructuredLogger.scala
morgen-peschke: I gave it a shot. To help avoid having to put this in every class header, I moved the documentation to it's own `README.md`

rossabaker (approved):
I like this a lot.  Useful with the right safety precautions, disastrous without, and a clear explanation which is which.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterLogger.scala
```diff
@@ -21,6 +21,16 @@ import cats.data._
 import cats.syntax.all._
 import org.typelevel.log4cats._
 
+/**
+ * A `SelfAwareLogger` implemented using `cats.data.Writer`.
+ *
+ * >>> WARNING: READ BEFORE USAGE! <<<
+ * https://github.com/typelevel/log4cats/blob/main/core/shared/src/main/scala/org/typelevel/log4cats/extras/README.md
```
danicheg, line 28: Wouldn't it be better to put these warnings on the microsite and leave here the link on it?

danicheg (approved):
Wow, such great PR is stuck for no reason! I dropped a minor question, so feel free to pass it over.

core/shared/src/main/scala/org/typelevel/log4cats/extras/WriterLogger.scala
morgen-peschke: IIRC neither of these classes is on the microsite, so putting the warning there might draw people to them, rather than nudge them towards less niche classes.

ChristopherDavenport (approved):

PR #720, opened 2023-01-10
add Natchez backend that provides a Logger[F] given a Trace[F]

bpholt:
I was looking at the [OpenTelemetry logging spec](https://opentelemetry.io/docs/reference/specification/logs/) the other day and saw this excerpt:

> This is in essence the philosophy behind OpenTelemetry’s logs support. We embrace existing logging solutions and make sure OpenTelemetry works nicely with existing logging libraries … 
> 
> For traces and metrics OpenTelemetry defines a new API that application developers must use to emit traces and metrics.
>
> For logs we did not take the same path. We realized that there is a much bigger and more diverse legacy in logging space. There are many existing logging libraries in different languages, each having their own API. Many programming languages have established standards for using particular logging libraries.

and it occurred to me that it might be nice to use the familiar log4cats `Logger[F]` interface for logging, even in apps that have Natchez `Trace[F]` wired throughout. 

I think this may also provide an easier migration path in an application that has existing `Logger[F]` statements throughout—one may be able to replace a `Logger` constraint with a `Trace` constraint, and then benefit from those existing log statements getting attached to spans going forward. (Eventually many of those logs will prove redundant and be removed.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,11 +119,11 @@ jobs:
 
       - name: Make target directories
         if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main')
-        run: mkdir -p testing/jvm/target noop/jvm/target target .js/target core/native/target site/target testing/native/target noop/native/target core/js/target js-console/target testing/js/target noop/js/target core/jvm/target .jvm/target .native/target slf4j/target project/target
+        run: mkdir -p testing/jvm/target noop/jvm/target natchez/jvm/target target .js/target core/native/target site/target testing/native/target natchez/js/target noop/native/target core/js/target js-console/target testing/js/target noop/js/target core/jvm/target .jvm/target .native/target slf4j/target natchez/native/target project/target
 
       - name: Compress target directories
         if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main')
-        run: tar cf targets.tar testing/jvm/target noop/jvm/target target .js/target core/native/target site/target testing/native/target noop/native/target core/js/target js-console/target testing/js/target noop/js/target core/jvm/target .jvm/target .native/target slf4j/target project/target
+        run: tar cf targets.tar testing/jvm/target noop/jvm/target natchez/jvm/target target .js/target core/native/target site/target testing/native/target natchez/js/target noop/native/target core/js/target js-console/target testing/js/target noop/js/target core/jvm/target .jvm/target .native/target slf4j/target natchez/native/target project/target
 
       - name: Upload target directories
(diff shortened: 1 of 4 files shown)


armanbilge:
👍 but this will create a circular dependency with natchez (which currently depends on log4cats). And also it would add an unstable 0.x dependency to log4cats which is in stable 2.x series.

Should we consider moving this either to Natchez repo, or creating a log4cats-natchez repo?

bpholt:
> this will create a circular dependency with natchez

I think this is only true for the `natchez-log` module, so it might be ok in practice?

armanbilge:
> I think this is only true for the `natchez-log` module, so it might be ok in practice?

Well, not really. Consider when breaking Scala Native 0.5 comes out, how will we upgrade these repos? Which will release first?

bpholt:
Oh, I see what you mean—I was thinking about it from a runtime perspective.

Maybe it should be a separate repo so it can be versioned independently from either side? (This isn't the place for the discussion, but I've been thinking that it might be a good idea to split out the Natchez backends generally, so I'm also hesitant to add more over there.)

Who could create a new typelevel repo for this? I'd be happy to set things up there, if someone wants to create the repo and make me the maintainer.

armanbilge:
> but I've been thinking that it might be a good idea to split out the Natchez backends generally

yup, I think so as well. honestly it needs its own GH org ...

--- 

> Who could create a new typelevel repo for this? I'd be happy to set things up there, if someone wants to create the repo and make me the maintainer.

I can do this, any preference on the name? :)

bpholt:
> I can do this, any preference on the name? :)

I proposed the module be `log4cats-natchez`, so that works for me.

We probably don't have to decide now, but that name raises some questions about the scope of the repo—e.g. it's not obvious _from the name alone_ that the log4cats backend for Natchez shouldn't belong there too. (I don't really want to tackle that right now, but I'd be open to discussing it if Natchez decides to move backends around)

armanbilge:
Done: https://github.com/typelevel/log4cats-natchez/

bpholt:
Great, thank you! I'll move this code over there and update this PR to mention the new backend in the log4cats documentation.

rossabaker (changes requested):
log4cats needs to have a very conservative strategy for breaking binary compatibility, given its uptake in several other projects.  We can't couple its release cycle to Natchez at its stage of maturity.  This integration could live in a separate repository, even hosted by Typelevel.

For a transparent integration of "legacy" logging and Open Telemetry, I would probably look first to a [log appender](https://opentelemetry.io/docs/reference/specification/logs/api/#how-to-create-log4j-style-appender).  For applications to generate events specifically with Otel in mind, there's an [event specification](https://opentelemetry.io/docs/reference/specification/logs/event-api/) that adds some extra conventional attributes.

bpholt:
I'm going to close this for now. I'll update the docs with a link to the new repo once it's published!

PR #738, opened 2023-03-24
Add strict evaluation option

### LINKED ISSUE #272: How to raise test coverage on parameters passed to NoOpLogger
I'm failing to meet 100% coverage requirements in a module because the call by name string inside a log effect never gets executed. Would it be reasonable to make the no-op logger execute the passed in strings or to provide an option that allowed it?

ant8e:
Fix #272

### DIFF
--- a/noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
+++ b/noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
@@ -22,13 +22,19 @@ import cats.Applicative
 object NoOpFactory extends LoggerFactoryGenCompanion {
   def apply[F[_]: Applicative]: LoggerFactory[F] = impl[F]
 
-  def impl[F[_]](implicit F: Applicative[F]): LoggerFactory[F] = new LoggerFactory[F] {
-    override def getLoggerFromName(name: String): SelfAwareStructuredLogger[F] = {
-      val _ = name
-      NoOpLogger.impl[F]
+  def strictEvalArgs[F[_]: Applicative]: LoggerFactory[F] = impl_[F](evaluateArgs = true)
+
+  def impl[F[_]](implicit F: Applicative[F]): LoggerFactory[F] = impl_[F](evaluateArgs = false)
+
+  private def impl_[F[_]](evaluateArgs: Boolean)(implicit F: Applicative[F]): LoggerFactory[F] =
+    new LoggerFactory[F] {
+      override def getLoggerFromName(name: String): SelfAwareStructuredLogger[F] = {
+        val _ = name
+        NoOpLogger.impl_[F](evaluateArgs)
+      }
+
+      override def fromName(name: String): F[SelfAwareStructuredLogger[F]] =
+        F.pure(getLoggerFromName(name))
     }
 
-    override def fromName(name: String): F[SelfAwareStructuredLogger[F]] =
-      F.pure(getLoggerFromName(name))
-  }
 }
--- a/noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
+++ b/noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
@@ -28,44 +28,83 @@ object NoOpLogger {
   def impl[F[_]: Applicative] = impl_[F](evaluateArgs = false)
 
   private[noop] def impl_[F[_]: Applicative](evaluateArgs: Boolean): SelfAwareStructuredLogger[F] =
-    new SelfAwareStructuredLogger[F] {
+    if (evaluateArgs) stricImpl else lazyImpl
 
-      val no: F[Boolean] = Applicative[F].pure(false)
-      def void(arg: => Any): F[Unit] = Applicative[F].pure {
-        if (evaluateArgs) {
-          val _ = arg
-          ()
-        } else ()
-      }
-      @inline override def isTraceEnabled: F[Boolean] = no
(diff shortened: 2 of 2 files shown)


noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
```diff
@@ -21,39 +21,51 @@ import org.typelevel.log4cats.SelfAwareStructuredLogger
...
+  def strictEvalArgs[F[_]: Applicative]: SelfAwareStructuredLogger[F] =
+    impl_[F](evaluateArgs = true)
+
+  def impl[F[_]: Applicative] = impl_[F](evaluateArgs = false)
+
+  private[noop] def impl_[F[_]: Applicative](evaluateArgs: Boolean): SelfAwareStructuredLogger[F] =
+    new SelfAwareStructuredLogger[F] {
+
+      val no: F[Boolean] = Applicative[F].pure(false)
+      def void(arg: => Any): F[Unit] = Applicative[F].pure {
```
rossabaker, line 34: What if we did the strictness check outside, and returned either an unconditionally strict or unconditionally lazy implementation?  Then we could avoid the conditional on each call (probably insignificant).  More importantly, we'd keep the F.unit optimization in the lazy case.

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
```diff
@@ -22,13 +22,19 @@ import cats.Applicative
 object NoOpFactory extends LoggerFactoryGenCompanion {
   def apply[F[_]: Applicative]: LoggerFactory[F] = impl[F]
 
-  def impl[F[_]](implicit F: Applicative[F]): LoggerFactory[F] = new LoggerFactory[F] {
-    override def getLoggerFromName(name: String): SelfAwareStructuredLogger[F] = {
-      val _ = name
-      NoOpLogger.impl[F]
+  def strictEvalArgs[F[_]: Applicative]: LoggerFactory[F] = impl_[F](evaluateArgs = true)
```
rossabaker, line 25: How are people planning to choose this?  An effectful toggle on a CI flag?

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
ant8e: Good point. Will do that

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
ant8e: As for me, the code under test receives an implicit  `Logger` or `LoggerFactory` parameter. So it is the test setup code that would create that `strictEvalArgs` instance.

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
```diff
@@ -28,44 +28,83 @@ object NoOpLogger {
...
+    def void(arg: => Any): F[Unit] = Applicative[F].pure {
+      val _ = arg
+      ()
     }
+
+    @inline override def isTraceEnabled: F[Boolean] = no
+    @inline override def isDebugEnabled: F[Boolean] = no
+    @inline override def isInfoEnabled: F[Boolean] = no
+    @inline override def isWarnEnabled: F[Boolean] = no
+    @inline override def isErrorEnabled: F[Boolean] = no
```
ant8e, line 82: The main use case of the `strictImpl` is to be able to exercise a much as possible code paths under test. So I'm wondering that maybe we should say yes instead here, in case the client code tests this before doing stuff

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpFactory.scala
danicheg: If this noop-things, why don't make them a default option?

ant8e: > If this noop-things, why don't make them a default option?

I see 2 reasons not to do it:

- It would change the current behaviour. Existing user would surely do not like to see their trace message becoming systematically evaluated in production.
- NoOp  means doing nothing. Evaluating args deviate a bit from that and should not be the default IMO

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
```diff
@@ -28,44 +28,83 @@ object NoOpLogger {
...
+      unit
+    @inline override def info(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+      unit
+    @inline override def warn(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+      unit
+    @inline override def error(ctx: Map[String, String], t: Throwable)(msg: => String): F[Unit] =
+      unit
+  }
+
+  private def stricImpl[F[_]: Applicative] = new SelfAwareStructuredLogger[F] {
```
rossabaker, line 70: Typo: `strictImpl`

noop/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
rossabaker: That sounds reasonable to me.

rossabaker (approved):

Issue #440, opened 2021-05-06
Blocking Slf4j Logger Backer


ChristopherDavenport:
Not all Slf4j Backends are async. As a result we should provide a blocking consructor so that users of those backends don't do their blocking logging on their computer pool which is wasteful.

Noted from Discord.

Impl can be the exact same as current with `Sync[F].delay -> Sync[F].blocking`

chaitanyawaikar:
I can pick this up.

armanbilge:
> What about changing current implementation for using `Sync[F].suspend` with default for `Delay`

That sounds like a good solution.

> implement as a new method inside `Slf4jLogger`?

👍

diogocanut:
I think this one can be closed already, solved at #762 .

PR #762, opened 2023-06-14
Blocker Slf4j Logger Backer

diogocanut:
Solves #440 .

### DIFF
--- a/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
+++ b/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
@@ -51,5 +51,5 @@ object Slf4jLogger extends Slf4jLoggerCompat {
     Sync[F].delay(getLoggerFromSlf4j[F](logger))
 
   def fromBlockingSlf4j[F[_]: Sync](logger: JLogger): F[SelfAwareStructuredLogger[F]] =
-    Sync[F].blocking(getLoggerFromBlockingSlf4j[F](logger))
+    Sync[F].delay(getLoggerFromBlockingSlf4j[F](logger))
 }
--- a/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
+++ b/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
@@ -35,6 +35,9 @@ object Slf4jLogger extends Slf4jLoggerCompat {
   def getLoggerFromSlf4j[F[_]: Sync](logger: JLogger): SelfAwareStructuredLogger[F] =
     new Slf4jLoggerInternal.Slf4jLogger(logger)
 
+  def getLoggerFromBlockingSlf4j[F[_]: Sync](logger: JLogger): SelfAwareStructuredLogger[F] =
+    new Slf4jLoggerInternal.Slf4jLogger(logger, Sync.Type.Blocking)
+
   def create[F[_]: Sync](implicit name: LoggerName): F[SelfAwareStructuredLogger[F]] =
     Sync[F].delay(getLoggerFromName(name.value))
 
@@ -46,4 +49,7 @@ object Slf4jLogger extends Slf4jLoggerCompat {
 
   def fromSlf4j[F[_]: Sync](logger: JLogger): F[SelfAwareStructuredLogger[F]] =
     Sync[F].delay(getLoggerFromSlf4j[F](logger))
+
+  def fromBlockingSlf4j[F[_]: Sync](logger: JLogger): F[SelfAwareStructuredLogger[F]] =
+    Sync[F].blocking(getLoggerFromBlockingSlf4j[F](logger))
 }
--- a/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
+++ b/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
@@ -37,10 +37,11 @@ private[slf4j] object Slf4jLoggerInternal {
   private[this] def contextLog[F[_]](
       isEnabled: F[Boolean],
       ctx: Map[String, String],
-      logging: () => Unit
+      logging: () => Unit,
+      sync: Sync.Type
   )(implicit F: Sync[F]): F[Unit] = {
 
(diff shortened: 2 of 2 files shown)


slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
```diff
@@ -59,63 +60,64 @@ private[slf4j] object Slf4jLoggerInternal {
...
   }
 
-  final class Slf4jLogger[F[_]](val logger: JLogger)(implicit F: Sync[F])
-      extends SelfAwareStructuredLogger[F] {
-    override def isTraceEnabled: F[Boolean] = F.delay(logger.isTraceEnabled)
-    override def isDebugEnabled: F[Boolean] = F.delay(logger.isDebugEnabled)
-    override def isInfoEnabled: F[Boolean] = F.delay(logger.isInfoEnabled)
-    override def isWarnEnabled: F[Boolean] = F.delay(logger.isWarnEnabled)
-    override def isErrorEnabled: F[Boolean] = F.delay(logger.isErrorEnabled)
+  final class Slf4jLogger[F[_]](val logger: JLogger, sync: Sync.Type = Sync.Type.Delay)(implicit
```
danicheg, line 63: MiMa is arguing against this change https://github.com/typelevel/log4cats/actions/runs/5270003952/jobs/9948431964?pr=762#step:13:29
But since this is a package private class actually, we can workaround this warning like it's done here https://github.com/typelevel/log4cats/blob/main/core/shared/src/main/scala/org/typelevel/log4cats/PagingSelfAwareStructuredLogger.scala#L80

danicheg (commented):
Sorry for slowpoking on reviewing this actually nice PR!

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
diogocanut: Pushed some changes, let me know if it's make sense

danicheg:
Thanks for iterating over this. LGTM! But let's wait for a while if other maintainers want to put their two cents on this.

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
```diff
@@ -59,63 +60,71 @@ private[slf4j] object Slf4jLoggerInternal {
...
+    def this(logger: JLogger)(
+        F: Sync[F]
+    ) =
+      this(logger, Sync.Type.Delay)(F)
+
+    override def isTraceEnabled: F[Boolean] = F.suspend(sync)(logger.isTraceEnabled)
+    override def isDebugEnabled: F[Boolean] = F.suspend(sync)(logger.isDebugEnabled)
+    override def isInfoEnabled: F[Boolean] = F.suspend(sync)(logger.isInfoEnabled)
+    override def isWarnEnabled: F[Boolean] = F.suspend(sync)(logger.isWarnEnabled)
+    override def isErrorEnabled: F[Boolean] = F.suspend(sync)(logger.isErrorEnabled)
```
armanbilge, line 77: I wonder if these ones should be kept as `F.delay(...)`.

1. I understand that actually logging may be blocking, but I don't think I heard of a logger where checking the configuration is blocking.

2. I guess a client may check these to avoid doing some expensive work if logging is disabled. But if they are treated as blocking then checking these can be very expensive.

danicheg:
Hey @diogocanut! Sorry for pinging, but your PR is good and I'd like to merge it, but there is a reasonable suggest from @armanbilge. Could you glance at it? Otherwise, I can do a final polish if you don't mind. Thank you!

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
diogocanut: Changed this to use `F.delay`. Thanks.

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
```diff
@@ -37,10 +37,11 @@ private[slf4j] object Slf4jLoggerInternal {
   private[this] def contextLog[F[_]](
       isEnabled: F[Boolean],
       ctx: Map[String, String],
-      logging: () => Unit
+      logging: () => Unit,
+      sync: Sync.Type
   )(implicit F: Sync[F]): F[Unit] = {
 
-    val ifEnabled = F.delay {
+    val ifEnabled = F.suspend(sync) {
```
armanbilge, line 44: Sorry, I think this is another one that should stay as `delay`. Interacting with the `MDC` should not be blocking, it's just a thread-local mutable map.

diogocanut: Done. Let me know if there is something else.

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
```diff
@@ -46,4 +49,7 @@ object Slf4jLogger extends Slf4jLoggerCompat {
 
   def fromSlf4j[F[_]: Sync](logger: JLogger): F[SelfAwareStructuredLogger[F]] =
     Sync[F].delay(getLoggerFromSlf4j[F](logger))
+
+  def fromBlockingSlf4j[F[_]: Sync](logger: JLogger): F[SelfAwareStructuredLogger[F]] =
+    Sync[F].blocking(getLoggerFromBlockingSlf4j[F](logger))
```
armanbilge, line 54: Sorry 😂 can this one be `delay` as well? `getLoggerFromBlockingSlf4j` is just calling `new` right? To make an instance of our own class. Making a new instance of a class should not be blocking right?

diogocanut: Done xd

armanbilge (approved):
thanks!

danicheg:
Great work on collaborating and making a valuable contribution 👍🏻

PR #778, opened 2023-08-16
Move `NoOpLogger` to core module

armanbilge:
Because `noop` module depends on `core` the final binary interface is identical. So, this is a compatible change, and we will keep enforcing that compatibility going forwards.


Daenyth (approved):
Fantastic change

danicheg (approved):
Giving strength and legitimacy to the approvals above

Issue #461, opened 2021-06-30
FiberLocal / IOLocal logger wrapper


kubukoz:
hopefully self-explanatory: it would be great to have a wrapper for loggers that would allow one to pass structured logging context in a more implicit fashion.

This would probably involve creating a root logging context (the default value of the IOLocal), then using that in a wrapped logger so it passes the context to the underlying logger (e.g. slf4j).

wunderk1nd-e:
Would something like this be helpful?
https://github.com/typelevel/log4cats/pull/676

kubukoz:
Yes, that's pretty much what I was looking for :)

Maybe some utility that would create the IOLocal for you would also be nice.

morgen-peschke:
I want to flag that we currently have 3 PRs working on this functionality (or functionality very similar to it), and we should probably reconcile them so we get the best result:

- #676 
- #900
- #905

kubukoz:
#676 can be closed, #905 is a direct continuation. #900, on the other hand...

Issue #525, opened 2021-11-18
Proposal for `log4cats-sourcecode` module to provide information about the call site


bplommer:
I suggest adding a modlue for integration with [sourcecode](https://github.com/com-lihaoyi/sourcecode) to add information about the call site to the context of a structured logger. Rather than introduce a new subtype of `Logger`, I think the simplest thing is to put the logging methods on an object. It might look something like this:

```scala
object SourceCodeLogger {
  def info[F[_]](ctx: Map[String, String])(msg: => String)(implicit logger: StructuredLogger, line: sourcecode.Line, file: sourcecode.File): F[Unit] =
    logger.info(ctx ++ Map("callsite" -> s"${file.value}:${line.value}"))(msg)
  /* ... */
}
```

Of course there's endless scope for bike shedding about exactly what information should be provided and in what format, so I don't know if this is over-opinionated to go in a library like log4cats?

ChristopherDavenport:
I think it could, although I'm not 100% on sourcecode as I havne't used it. I normally use

https://github.com/tpolecat/SourcePos

for this sort of thing, but can be convinced for sure.

bplommer:
sourcepos looks good! I like that it only needs a single parameter and has its own opinionated toString.

darkfrog26:
Scribe uses `sourcecode` since 3.x and it works beautifully with `scribe-cats`.

PR #800, opened 2023-10-26
Allow adding effect's contextual environment to the log context

iRevive:
Closes https://github.com/typelevel/log4cats/issues/461. Some ideas are based on #676.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -64,7 +64,6 @@ lazy val mtl = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     name := "log4cats-mtl",
     libraryDependencies += "org.typelevel" %%% "cats-mtl" % catsMtlV
   )
-  .nativeSettings(commonNativeSettings)
 
 lazy val testing = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(commonSettings)


build.sbt
```diff
@@ -47,7 +48,8 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     name := "log4cats-core",
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-core"       % catsV,
-      "org.typelevel" %%% "cats-effect-std" % catsEffectV
+      "org.typelevel" %%% "cats-effect-std" % catsEffectV,
+      "org.typelevel" %%% "cats-mtl"        % catsMtlV
```
iRevive, line 52: The drawback of having this functionality in the `core` module. 

I can move this code into `log4cats-extras` module. It could be handy if we want to provide some 'non-trivial' functionality: invocation positions (#525), etc.

danicheg: Just an idea, maybe it's worth creating a `log4cats-mtl` module then?

iRevive: Valid point

iRevive:
I made `log4cats-mtl` module.

Should I set an upcoming release via `tlVersionIntroduced := List("2.12", "2.13", "3").map(_ -> "2.7.0").toMap` to suppress MiMa errors in the module? Or should I bump the base version instead: `tlBaseVersion := 2.7`?

armanbilge:
If we are making a `log4cats-mtl` module, we should consider that `cats-mtl` may become a direct dependency of Cats Effect via https://github.com/typelevel/cats-effect/issues/3385. Although that might be Cats Effect core only, and not the `std` module we're using here.

Just want to make sure that we think about the migration path in the case MTL ends up on the `log4cats-core` classpath (an acceptable migration path may be to accept the less-than-ideal situation as a historical artifact :)

iRevive:
Is there any ETA? It would be nice to have Local as a part of cats-effect.

lenguyenthanh:
`cats-mtl` is now part of `cats-effect`, should we revive (pun intended) this one or wait for the bigger fish in #920 ?

iRevive:
I think we should wait for #920 first.

PR #825, opened 2024-03-27
Quality of life updates for LoggerFactory

morgen-peschke:
Adds methods to return updated versions that configure the loggers they produce using `addContext` and `withModifiedString`.

This should make it easier to migrate from passing around a single `SelfAwareStructuredLogger` and using a `LoggerFactory`.

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/LoggerFactory.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/LoggerFactory.scala
@@ -17,6 +17,7 @@
 package org.typelevel.log4cats
 
 import cats.Functor
+import cats.Show.Shown
 import cats.data.Kleisli
 import cats.syntax.functor._
 import cats.~>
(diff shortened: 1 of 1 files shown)


rossabaker (approved):
I don't know what the `LoggerFactoryGen` supertype is for, but I'll ask: would this have worked on `LoggerFactoryGen`?  I think probably not, because then it wouldn't be able to return a `LoggerFactory`?

Assuming that's a bad idea, :+1:

morgen-peschke:
> I don't know what the `LoggerFactoryGen` supertype is for, but I'll ask: would this have worked on `LoggerFactoryGen`? I think probably not, because then it wouldn't be able to return a `LoggerFactory`?
> 
> Assuming that's a bad idea, 👍

I'm pretty sure that it couldn't go on `LoggerFactoryGen` because they get passed through to methods which aren't on `Logger`, so we need the constraint that `LoggerFactory` adds that `LoggerFactoryGen#LoggerType` is `SelfAwareStructuredLogger[F]` instead of any `Logger[F]` subtype.

PR #845, opened 2024-05-29
Add `.sbtopts`

### LINKED ISSUE #844: OOME occurs on CI when building library against Scala 3 on Native
> `Error: Exception in thread "classloader-cache-cleanup-0" java.lang.OutOfMemoryError: GC overhead limit exceeded`

https://github.com/typelevel/log4cats/actions/runs/9282306444/job/25540033178#step:10:86

danicheg:
Fixes #844

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,6 @@
+-J-Dfile.encoding=UTF-8
+-J-Xms1G
+-J-Xmx6G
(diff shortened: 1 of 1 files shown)


danicheg:
One incidental but pleasant outcome of this is that the `nativeLink` step on CI now takes 2x times less time to proceed.

PR #850, opened 2024-06-14
Fix MDC handling in Slf4jLogger

morgen-peschke:
Updates the handling of the MDC in `Slf4jLogger` to resolve questions raised in #851 

Changes:
- Always clear the MDC before populating when logging with context
- Don't skip the MDC massaging when logging without context

Checks:
- laziness of `msg` parameter
- dispatch to log4j Logger messages
- MDC setup/restoration
- MDC recovery when exceptions happen during message creation

PR #854, opened 2024-06-17
Improve API around transforming loggers

morgen-peschke:
Inspired by the tedium of #835

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/ErrorLogger.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/ErrorLogger.scala
@@ -16,13 +16,17 @@
 
 package org.typelevel.log4cats
 
-import cats._
+import cats.*
+import org.typelevel.log4cats.extras.LogLevel
 trait ErrorLogger[F[_]] {
-  def error(t: Throwable)(message: => String): F[Unit]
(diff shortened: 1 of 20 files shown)


rossabaker:
I have imagined something like this, but it does result in a switch statement on every logging operation, at least for the common slf4j backend.  Does anybody have an inkling of the performance impact of this?

morgen-peschke:
> I have imagined something like this, but it does result in a switch statement on every logging operation, at least for the common slf4j backend. Does anybody have an inkling of the performance impact of this?

I'm not sure, though I'm optimistic.

Since there's already at least 1-2 conditionals per logging operation (checking the log level, and if to restore or clear the MDC, in addition to whatever goes on inside the MDC and Log4J) my gut-check is it won't be enough to move the needle 🤞🏻

I don't think we have timing checks on the log messages, though #850 would be a great place to add those (not sure what that would look like, how many messages do we expect to be able to reliable log in a tight loop within `N` seconds?)

rossabaker:
Yeah, any performance metrics here risk falling into the same trap we see in the `IO` world: the speed of successive flatMaps can matter while the `IO` is trivial, but the speed of a gajillion flatMaps is insignificant in the vastness of a real network IO.

PR #890, opened 2025-02-13
Fix some warnings related to -xsource:3

### LINKED ISSUE #889: Fix style inconsistencies
> I was a little surprised that the imports can be mixed between `*` and `_` styles. Might be good to follow up to switch entirely to `*` for consistency, but absolutely not blocking this PR.

_Originally posted by @samspills in https://github.com/typelevel/log4cats/pull/850#pullrequestreview-2599411811_

morgen-peschke:
Fixes #889

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,33 +41,33 @@ Good news, you don't have to!  Enter log4cats!  Read on!
 import org.typelevel.log4cats.Logger
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 import cats.effect.Sync
-import cats.implicits._
+import cats.implicits.*
 
 object MyThing {
   // Impure But What 90% of Folks I know do with log4s
-  implicit def unsafeLogger[F[_]: Sync] = Slf4jLogger.getLogger[F]
+  implicit def unsafeLogger[F[_] : Sync] = Slf4jLogger.getLogger[F]
 
   // Arbitrary Local Function Declaration
-  def doSomething[F[_]: Sync]: F[Unit] =
+  def doSomething[F[_] : Sync]: F[Unit] =
     Logger[F].info("Logging Start Something") *>
-    Sync[F].delay(println("I could be doing anything"))
-      .attempt.flatMap{
-        case Left(e) => Logger[F].error(e)("Something Went Wrong")
-        case Right(_) => Sync[F].pure(())
-      }
+            Sync[F].delay(println("I could be doing anything"))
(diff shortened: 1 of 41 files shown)


README.md
```diff
@@ -41,33 +41,33 @@ Good news, you don't have to!  Enter log4cats!  Read on!
 import org.typelevel.log4cats.Logger
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 import cats.effect.Sync
-import cats.implicits._
+import cats.implicits.*
 
 object MyThing {
   // Impure But What 90% of Folks I know do with log4s
-  implicit def unsafeLogger[F[_]: Sync] = Slf4jLogger.getLogger[F]
+  implicit def unsafeLogger[F[_] : Sync] = Slf4jLogger.getLogger[F]
```
rossabaker, line 48: Is there a warning on these spaces?  This is the thing that seems strangest to me.

rossabaker (approved):
I admit I've lost all track of what syntax is in and out of fashion anymore.  The build passes and I'm not offended by anything, so, ✅

README.md
bcarter97: The default `spaces.beforeContextBoundColon` is [Never](https://scalameta.org/scalafmt/docs/configuration.html#spacesbeforecontextboundcolon) and it doesn't seem [configured](https://github.com/typelevel/log4cats/blob/main/.scalafmt.conf) so it's a bit weird these have been added

README.md
```diff
@@ -41,33 +41,33 @@ Good news, you don't have to!  Enter log4cats!  Read on!
 import org.typelevel.log4cats.Logger
 import org.typelevel.log4cats.slf4j.Slf4jLogger
 import cats.effect.Sync
-import cats.implicits._
+import cats.implicits.*
```
bcarter97, line 44: Should these all be changed to `cats.syntax.all.*`? It's mentioned in these release notes that the latter is preferred
https://github.com/typelevel/cats/releases/tag/v2.2.0

rossabaker: I'm indifferent to whether it happens in this PR, but :+1: in general.

morgen-peschke: It's only in 2 places, so it's worth doing now rather than dealing with the overhead of another PR

README.md
morgen-peschke: I'm going to assume that it's something weird with my editor, so I've rolled it back.

samspills (approved):

PR #900, opened 2025-03-14
Adds a StructuredLogger based on cats.mtl.Ask

rossabaker:


rossabaker:
The flake.* files are not germane to this PR, and could be spun off into a different one.  They were germane to me getting a working Metals.

build.sbt
```diff
@@ -47,7 +48,8 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     name := "log4cats-core",
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-core"       % catsV,
-      "org.typelevel" %%% "cats-effect-std" % catsEffectV
+      "org.typelevel" %%% "cats-effect-std" % catsEffectV,
+      "org.typelevel" %%% "cats-mtl"        % catsMtlV
```
rossabaker, line 52: This dependency will be part of cats-effect-3.6, so I'm not feeling guilty for introducing it to core.  Otherwise, this could be a log4cats-mtl module.

core/shared/src/main/scala/org/typelevel/log4cats/AskStructuredLogger.scala
```diff
@@ -0,0 +1,71 @@
...
+    F.flatMap(ask.ask)(askCtx => logger.info(askCtx, t)(message))
+
+  override def debug(t: Throwable)(message: => String): F[Unit] =
+    F.flatMap(ask.ask)(askCtx => logger.debug(askCtx, t)(message))
+
+  override def trace(t: Throwable)(message: => String): F[Unit] =
+    F.flatMap(ask.ask)(askCtx => logger.trace(askCtx, t)(message))
+
+  override def trace(ctx: Map[String, String])(msg: => String): F[Unit] =
+    F.flatMap(ask.ask)(askCtx => logger.trace(askCtx ++ ctx)(msg))
```
rossabaker, line 42: Performance concerns, not yet proven by benchmarking:

1. We pay to ask and flatMap when the level is disabled.
2. In the methods with a `ctxt`, we additionally pay to concatenate askCtx and ctx when the level is disabled.

An underlying SelfAwareLogger could help, but then we'd be an AskSelfAwareStructuredLogger.  The traits aren't composing particularly well.

We could further optimize by getting into `Slf4jLogger`, but that may further contribute to the combinatorial explosion.

core/shared/src/main/scala/org/typelevel/log4cats/AskStructuredLogger.scala
```diff
@@ -0,0 +1,71 @@
+package org.typelevel.log4cats
+
+import cats.Monad
+import cats.mtl.Ask
+
+private[log4cats] class AskStructuredLogger[F[_]](logger: StructuredLogger[F])(implicit
+    F: Monad[F],
+    ask: Ask[F, Map[String, String]]
```
rossabaker, line 8: The `Map` returned here could be a view (say, from otel4s), but it needs to have a fast `++` to another `Map`.

core/shared/src/main/scala/org/typelevel/log4cats/AskStructuredLogger.scala
NthPortal: as annoying as it would be, I don't think a single "duplicate" for `SelfAwareStructuredLogger` is *too* awful

core/shared/src/main/scala/org/typelevel/log4cats/AskStructuredLogger.scala
```diff
@@ -0,0 +1,71 @@
+package org.typelevel.log4cats
+
+import cats.Monad
+import cats.mtl.Ask
+
+private[log4cats] class AskStructuredLogger[F[_]](logger: StructuredLogger[F])(implicit
```
NthPortal, line 6: I know it's out of character for this repo, but could you add some docs and tests?

rossabaker: Definitely.  I wanted to iterate on the approach a little bit first before investing too much in that.

This should work for non-slf4j loggers, with the aforementioned performance concerns.  Is it worth trying to do both, or do we think it's better to focus on the slf4j that everyone uses in practice?

NthPortal: I don't think the performance can be meaningfully worse than the existing terrible performance of every call to `addContext` adding another `Map` concatenation for *every log operation*

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/Slf4jLogger.scala
```diff
@@ -33,10 +34,10 @@ object Slf4jLogger extends Slf4jLoggerCompat {
     getLoggerFromSlf4j[F](org.slf4j.LoggerFactory.getLogger(clazz))
 
   def getLoggerFromSlf4j[F[_]: Sync](logger: JLogger): SelfAwareStructuredLogger[F] =
-    new Slf4jLoggerInternal.Slf4jLogger(logger, Sync.Type.Delay)
+    new Slf4jLoggerInternal.Slf4jLogger(logger, Sync.Type.Delay, Applicative[F].pure(Map.empty))
```
rossabaker, line 37: This (and line 40) wire nothing in yet.  We'd need a parallel set of Ask or Local based constructors here.

slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
```diff
@@ -77,14 +73,19 @@ private[slf4j] object Slf4jLoggerInternal {
   }
 
   @nowarn("msg=used")
-  final class Slf4jLogger[F[_]](val logger: JLogger, sync: Sync.Type = Sync.Type.Delay)(implicit
-      F: Sync[F]
-  ) extends SelfAwareStructuredLogger[F] {
+  final class Slf4jLogger[F[_]](
+      val logger: JLogger,
+      sync: Sync.Type = Sync.Type.Delay,
+      defaultCtx: F[Map[String, String]]
```
rossabaker, line 79: By getting the default context into the logger, as an effect, we are now able to avoid the asking and concatenation of the context map when the log is disabled.  But this is a bit less conducive to the middleware approach: we don't want `addContext` to return a `ModifiedContextStructuredLogger` anymore.

PR #902, opened 2025-03-18
Avoid tupling when putting to MDC

rossabaker:
This can avoid creating tuples that will immediately be thrown away by certain kinds of maps.


rossabaker:
I am receptive to being thwacked with the "not without a benchmark" stick.

rossabaker:
Looks like that's no good in 2.12 either.

Issue #868, opened 2024-10-27
Unable to use turbo filters dependent on MDC


r-glyde:
We're trying to use [logback turbofilters](https://logback.qos.ch/manual/filters.html#TurboFilter) to alter which logs are output based on values in the MDC but this doesn't seem to work properly with log4cats and the slf4j logger implementation. When outputting the MDC as part of the log I can see it includes the correct value but the correct filtering doesn't get applied when trying to output logs below the global log level (e.g. log at `DEBUG`, set global level to `INFO` but force `ALLOW` in the filter). This is with both the `MDCFilter` and the `DynamicThresholdFilter`.

This appears to be because in `Slf4jLoggerInternal`, whenever a logging method is called there is a check of the corresponding `is$levelEnabled` value which doesn't include updating the MDC in the versions where a context is passed in (e.g. for debug [here](https://github.com/typelevel/log4cats/blob/a4191f963812ed71158d3c84ca9863998e8dc9ea/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala#L92-L93)).

I can't think of a reason we wouldn't want these filters to work (happy to be corrected) so was thinking that in [contextLog](https://github.com/typelevel/log4cats/blob/a4191f963812ed71158d3c84ca9863998e8dc9ea/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala#L37-L60) the MDC should be updated prior to the `isEnabled` check? Alternative could be to not run the `is$levelEnabled` checks prior to each log and leave to the user but probably nice that that safety of avoiding running anything expensive if it won't be logged is built in? I've tried these locally (roughly) and they both seem to work as expected. Happy to look at raising a PR for either of these if it makes sense.

Semi-related - in some projects, we're using a custom self aware logger implementation that wraps an underlying logger where the overridden `is$levelEnabled` methods match what would be enabled in the turbo filter but this is then not used in the end because the check is done within `Slf4jLoggerInternal` using it's own implementation. I'm not sure if there is anything we can do to avoid this (without some more significant changes to how the logging is implemented) but ends up being a bit confusing that we have two implementations of the method and ours isn't the actual one being used when the check is done?

bcarter97:
@rossabaker @danicheg this would be really useful for us, are you ok with this being raised as a PR to discuss?

danicheg:
[...]

PR #907, opened 2025-04-05
Reimplement contextLog to populate MDC even when disabled

bcarter97:
My attempt at solving #868. 

- Populate the MDC even if isEnabled resolves in `false`
  - This is tested by implementing a TestLogger that will update some state with the current MDC value every time the `<level>isEnabled` is checked, to assert it is updated even if the check resolves to false
- Changed the error handling to be more functional
  - not sure if this will impact performance? It will result in a few more flatMaps.

### DIFF
--- a/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
+++ b/slf4j/src/main/scala/org/typelevel/log4cats/slf4j/internal/Slf4jLoggerInternal.scala
@@ -16,12 +16,13 @@
 
 package org.typelevel.log4cats.slf4j.internal
 
-import org.typelevel.log4cats.*
-import cats.syntax.all.*
 import cats.effect.*
-import org.slf4j.Logger as JLogger
-import org.slf4j.MDC
+import cats.effect.syntax.all.*
+import cats.syntax.all.*
+import org.slf4j.{Logger as JLogger, MDC}
+import org.typelevel.log4cats.*
 
+import java.util
 import scala.annotation.nowarn
 
 private[slf4j] object Slf4jLoggerInternal {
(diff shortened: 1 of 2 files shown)


danicheg:
@morgen-peschke Would you mind reviewing this PR?

bcarter97:
> > not sure if this will impact performance? It will result in a few more flatMaps.
> 
> Took a minute to confirm my suspicion here, but the concern isn't the performance, it's that the `MDC` is a thread-local construct and multiple fibers can share the same thread. Manipulating the `MD` needs to be done within a single `Sync[F].delay` block to prevent multiple loggers from clobbering the `MDC`.
> 
> Something like this would be trouble:
> 
>     1. Logger A clears the `MDC` and yields
> 
>     2. Logger B clears the `MDC` and yields
> 
>     3. Logger A populates the `MDC` and yields
> 
>     4. Logger B populates the `MDC` and yields
> 
>     5. Logger A logs a message with the combined context and yields
> 
>     6. Logger A resets the `MDC` and yields
> 
>     7. Logger B logs a message without it's context and yields
> 
>     8. Logger B resets the `MDC` and yields

Riight I see, that makes sense. I am now a bit unsure how to implement this in one effect, as I think we need:

1. Set MDC
2. Do log
3. backup MDC

This worked before because isEnabled was done as step one. But to properly set the MDC before the log is called (which is needed for TurboFilters to work) it has to be second. Do you have any ideas on how this could work? I will continue to have a look into this as well.

armanbilge:
> Do you have any ideas on how this could work? I will continue to have a look into this as well.

@bcarter97 here's a possible idea you can explore. In Cats Effect v3.6.0, we introduced the ability to create a `ThreadLocal` "view" that is backed by an `IOLocal`. This guarantees that the `ThreadLocal` behaves consistently across `flatMap`s even as `Fiber`s move between threads.

https://github.com/typelevel/cats-effect/blob/553fdbbe62349a0d059213ed7af099db2cd98af4/core/jvm/src/main/scala/cats/effect/IOLocalPlatform.scala#L23-L29

The part I'm less sure about, that you should explore, is if it's possible to make a custom MDC implementation that is based on the `IOLocal`-backed `ThreadLocal`. We were able to do this in otel4s.

- https://github.com/typelevel/otel4s/pull/214

morgen-peschke:
I think that, with a bit of restructuring, we can get there without major structural changes to `Sl4jLoggerInternal`.

I was poking at it and ended up with #910, and it looks like it might do what you need.

bcarter97:
> I think that, with a bit of restructuring, we can get there without major structural changes outside of `Sl4jLoggerInternal`.
> 
> I was poking at it and ended up with #910, and it looks like it might do what you need.

Yeah I was doing some digging and I don't think it will be possible with only 1 effect without exposing the logger's `isEnabled` check.

@armanbilge I had a look at how this could be done with IOLocal - slf4j exposes an SPI for MDCAdapter. I've looked at how Otel4s does this with IOLocalContextStorage, but I can't say I'm very confident I'd be able to replicate something similar.

I think I'll close this PR in favour of #910, as the implementation is flawed.

PR #909, opened 2025-04-09
Add logger with `Local` semantics

NthPortal:
- [ ] docs
- [ ] tests
- [ ] remove otel4s dependency (see typelevel/cats-mtl#625)

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/LocalLogContext.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/LocalLogContext.scala
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2018 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/org/typelevel/log4cats/LocalLogContext.scala
```diff
@@ -0,0 +1,127 @@
...
+  private[log4cats] def withAddedContext[A](ctx: Map[String, String])(fa: F[A]): F[A]
+
+  private[log4cats] final def withAddedContext[A](ctx: (String, Show.Shown)*)(fa: F[A]): F[A] =
+    withAddedContext {
+      ctx.view.map { case (k, v) => k -> v.toString }.toMap
+    }(fa)
+
+  def withHighPriorityAskedContext(ask: Ask[F, Map[String, String]]): LocalLogContext[F]
+
+  def withLowPriorityAskedContext(ask: Ask[F, Map[String, String]]): LocalLogContext[F]
```
NthPortal, line 39: these allow you to, for example, add trace and span IDs to the context. high/low priority is which wins if there's a key collision (hopefully never happens for span context, but could for other things)

NthPortal:
depends on #924 for cats-mtl 1.6.0 (which only exists for native 0.5.x)

PR #910, opened 2025-04-10
Expose MDC to Slf4jLoggerInternal isEnabled

morgen-peschke:
One approach to enable functionality requested in #868 (see also discussion in #907)

> [!NOTE]
> This will mean doing `MCD` manipulation every time `isEnabled` is called, but I don't really see a way around that if we want to support this functionality.


bcarter97:
@danicheg any thoughts on this? Would be a very useful addition for us 😄

rossabaker (approved):

PR #920, opened 2025-06-18
SAM Type Interface

Jay-Lokhande:
This PR's objective is to create a SAM type interface in log4cats

### DIFF
--- a/core/shared/src/main/scala/org/typelevel/log4cats/KernelLogLevel.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/KernelLogLevel.scala
@@ -26,7 +26,9 @@ sealed abstract class KernelLogLevel(val name: String, val value: Int) {
 }
 
 object KernelLogLevel {
-  implicit final val log4catsCatsInstances: Eq[KernelLogLevel] & Hash[KernelLogLevel] & Order[KernelLogLevel] & Show[KernelLogLevel] =
+  implicit final val log4catsCatsInstances: Eq[KernelLogLevel] & Hash[KernelLogLevel] & Order[
+    KernelLogLevel
+  ] & Show[KernelLogLevel] =
     new Hash[KernelLogLevel] with Order[KernelLogLevel] with Show[KernelLogLevel] {
       override def eqv(x: KernelLogLevel, y: KernelLogLevel): Boolean = x == y
 
--- a/core/shared/src/main/scala/org/typelevel/log4cats/Log.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/Log.scala
@@ -205,7 +205,9 @@ object Log {
 
     override def withThrowable(throwable: Throwable): Builder[Ctx] = this
 
-    override def withContext[A](name: String)(ctx: A)(implicit E: Context.Encoder[A, Ctx]): Builder[Ctx] = this
+    override def withContext[A](name: String)(ctx: A)(implicit
+        E: Context.Encoder[A, Ctx]
+    ): Builder[Ctx] = this
 
     override def withFileName(name: String): Builder[Ctx] = this
 
--- a/core/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLoggerKernel.scala
+++ b/core/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLoggerKernel.scala
@@ -26,7 +26,10 @@ object NoOpLoggerKernel {
   }
 
   def strict[F[_]: Applicative, Ctx]: LoggerKernel[F, Ctx] = new LoggerKernel[F, Ctx] {
-    override def log(level: KernelLogLevel, record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] = {
+    override def log(
+        level: KernelLogLevel,
+        record: Log.Builder[Ctx] => Log.Builder[Ctx]
+    ): F[Unit] = {
       val _ = record(Log.strictNoOpBuilder[Ctx]())
       Applicative[F].unit
     }


core/shared/src/main/scala/org/typelevel/log4cats/LoggerKernel.scala
```diff
@@ -0,0 +1,49 @@
...
+  final def logTrace(record: Log.Builder => Log.Builder): F[Unit] =
+    log(KernelLogLevel.Trace, record)
+  final def logDebug(record: Log.Builder => Log.Builder): F[Unit] =
+    log(KernelLogLevel.Debug, record)
+  final def logInfo(record: Log.Builder => Log.Builder): F[Unit] =
+    log(KernelLogLevel.Info, record)
+  final def logWarn(record: Log.Builder => Log.Builder): F[Unit] =
+    log(KernelLogLevel.Warn, record)
+  final def logError(record: Log.Builder => Log.Builder): F[Unit] =
+    log(KernelLogLevel.Error, record)
```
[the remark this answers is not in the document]
Jay-Lokhande: lets keep it until final change, I tried to change as suggested but its giving error due to duplicates

core/shared/src/main/scala/org/typelevel/log4cats/KernelLogLevel.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  private var map = Map.empty[String, KernelLogLevel]
+  private var padded = Map.empty[KernelLogLevel, String]
+
+  implicit final val orderKernelLogLevel: Order[KernelLogLevel] =
+    Order.by[KernelLogLevel, Int](-_.value)
+
+  // For Java/legacy interop, if needed (not implicit)
+  val LevelOrdering: Ordering[KernelLogLevel] =
+    Ordering.by[KernelLogLevel, Int](_.value).reverse
```
morgen-peschke, line 38: Java interop is a non-issue, since attempting to use cats-effect from Java would be painful (at best) and the idioms are all sorts of incompatible.

There's an implicit which provides an `Ordering[A]` if there is an `Order[A]` in scope, so this can be removed.

core/shared/src/main/scala/org/typelevel/log4cats/Log.scala
```diff
@@ -45,112 +45,96 @@ object Log {
...
 
-  private class MutableBuilder[Ctx] private[Log] () extends Builder[Ctx] with Log[Ctx] {
+  private val noopMessage: () => String = () => ""
+
+  private class MutableBuilder[Ctx] extends Builder[Ctx] {
     private var _timestamp: Option[FiniteDuration] = None
     private var _level: Option[KernelLogLevel] = None
-    private var _levelValue: Option[Int] = None
-    private var _message: Option[String] = None
+    private var _message: () => String = noopMessage
```
morgen-peschke, line 71: Question: is the use case of a log without a message something we want to support?

core/shared/src/main/scala/org/typelevel/log4cats/Log.scala
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Low-level interface exposing methods to enrich a log record with relevant information. The
+ * methods are designed to capture elements that cannot be easily captured from a monadic context
+ * (or by running an effect). Elements such as timestamps should be provided by means of
+ * middlewares.
+ */
+trait Log[Ctx] {
+  def timestamp: Option[FiniteDuration]
+  def level: KernelLogLevel
+  def levelValue: Int
```
morgen-peschke, line 31: These should probably be collapsed to avoid a situation where they are out of sync.

core/shared/src/main/scala/org/typelevel/log4cats/SamLogger.scala
```diff
@@ -0,0 +1,154 @@
...
+  ): SamLogger[Kleisli[F, A, *], Ctx] =
+    ev.mapK(Kleisli.liftK[F, A])
+
+  private def withModifiedString[F[_], Ctx](
+      l: SamLogger[F, Ctx],
+      f: String => String
+  ): SamLogger[F, Ctx] =
+    new SamLogger[F, Ctx] {
+      def log(level: KernelLogLevel, record: Builder => Builder): F[Unit] = {
+        val modifiedRecord = (builder: Builder) => {
```
morgen-peschke, line 122: Needing to materialize the log and manually copy over the full message hints to me that we may need to refine `Builder`.

Pretty sure this would materialize `message` as well, so that's something to check.

core/shared/src/main/scala/org/typelevel/log4cats/Log.scala
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Low-level interface exposing methods to enrich a log record with relevant information. The
+ * methods are designed to capture elements that cannot be easily captured from a monadic context
+ * (or by running an effect). Elements such as timestamps should be provided by means of
+ * middlewares.
+ */
+trait Log[Ctx] {
+  def timestamp: Option[FiniteDuration]
+  def level: KernelLogLevel
+  def message: String
```
morgen-peschke, line 31: This should probably be lazy all the way to the edge.

```suggestion
  def message: () => String
```

core/shared/src/main/scala/org/typelevel/log4cats/SamLogger.scala
morgen-peschke: With the added `adaptMessage` method, this becomes:
```scala
      def log(level: KernelLogLevel, record: Builder => Builder): F[Unit] = {
        l.log(level, record(_).adaptMessage(f))
      }
```

core/shared/src/main/scala/org/typelevel/log4cats/SamStructuredLogger.scala
```diff
@@ -0,0 +1,326 @@
...
+trait SamStructuredLogger[F[_]] extends Logger[F] {
+  protected def kernel: LoggerKernel[F, String]
+
+  def trace(ctx: Map[String, String])(msg: => String): F[Unit] = {
+    val builder = (b: Log.Builder[String]) => {
+      var current = b.withMessage(msg)
+      ctx.foreach { case (k, v) => current = current.withContext[String](k)(v) }
+      current
+    }
+    kernel.logTrace(builder)
```
morgen-peschke, line 37: Might be worth exercising the methods on `Log.Builder` to see if they pull their weight. 
For example, this can be simplified to:
```suggestion
    kernel.logTrace(_.withMessage(msg).withContextMap(ctx))
```

core/shared/src/main/scala/org/typelevel/log4cats/LogRecord.scala
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.typelevel.log4cats
+
+/**
+ * This allows to capture several elements in vararg-based interface methods, enriching a single log
+ * with various pieces of information.
+ *
+ * This allows for an interesting UX, where the details of the encoding of some data into a log can
+ * be separate from the actual log statements.
```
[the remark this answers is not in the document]
morgen-peschke: Reminder to remove this

build.sbt
```diff
@@ -47,7 +48,8 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     name := "log4cats-core",
     libraryDependencies ++= Seq(
       "org.typelevel" %%% "cats-core"       % catsV,
-      "org.typelevel" %%% "cats-effect-std" % catsEffectV
+      "org.typelevel" %%% "cats-effect-std" % catsEffectV,
+      "com.lihaoyi" %%% "sourcecode"        % sourcecodeV
```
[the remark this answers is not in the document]
morgen-peschke: Reminder to remove the `sourcecode` macros until we can figure out the bin/source-compat story.

core/shared/src/main/scala/org/typelevel/log4cats/KernelLogLevel.scala
```diff
@@ -0,0 +1,59 @@
...
+    map += level.name.toLowerCase -> level
+    if (length > maxLength) {
+      maxLength = length
+      padded = map.map { case (_, level) =>
+        level -> level.name.padTo(maxLength, ' ').mkString
+      }
+    } else {
+      padded += level -> level.name.padTo(maxLength, ' ').mkString
+    }
+  }
```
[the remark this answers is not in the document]
morgen-peschke: Reminder to remove the dynamically added log levels

core/shared/src/main/scala/org/typelevel/log4cats/Log.scala
morgen-peschke: @kubukoz can you think of an reason we'd want to support a log with a message?

If not, this could probably be moved to a constructor parameter, and we could drop `withMessage`

core/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLoggerKernel.scala
```diff
@@ -0,0 +1,29 @@
...
+import org.typelevel.log4cats.{KernelLogLevel, Log, LoggerKernel}
+
+class NoOpLoggerKernel[F[_]: Applicative, Ctx] extends LoggerKernel[F, Ctx] {
+  def log(level: KernelLogLevel, record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] =
+    Applicative[F].unit
+}
+
+object NoOpLoggerKernel {
+  def apply[F[_]: Applicative, Ctx]: NoOpLoggerKernel[F, Ctx] = new NoOpLoggerKernel[F, Ctx]
+}
```
morgen-peschke, line 29: No need for a class, it's both simpler (and a good test of the SAM-ness of `LoggerKernel`) to implement it directly
```suggestion
object NoOpLoggerKernel {
  def apply[F[_]: Applicative, Ctx]: LoggerKernel[F, Ctx] = new LoggerKernel[F, Ctx] {
    override def log(level: KernelLogLevel, record: Builder => Builder): F[Unit] =
      Applicative[F].unit
  }
}
```

Interestingly, this does highlight that `LoggerKernel` is not currently a SAM type

core/shared/src/main/scala/org/typelevel/log4cats/LoggerKernel.scala
```diff
@@ -0,0 +1,51 @@
...
+ *     JSON library.
+ *   - The SAM-like nature of the construct makes it inherently middleware friendly, as a single
+ *     method call needs to be intercepted/proxied in order to amend the behaviour of the logger.
+ *
+ * This also means that different libraries can use wrappers on top of this kernel interface to use
+ * whatever UX is preferred without necessarily imposing constraints on the underlying
+ * implementation.
+ */
+trait LoggerKernel[F[_], Ctx] {
+  type Builder = Log.Builder[Ctx]
```
morgen-peschke, line 37: This turns out to be the offending line. If `Builder` is inlined, then it does become a SAM type.

(The difference is that `Builder` can be overridden, which breaks the SAM-ness of the trait)

core/shared/src/main/scala/org/typelevel/log4cats/LoggerKernel.scala
```diff
@@ -0,0 +1,51 @@
...
+  final def logTrace(record: Builder => Builder): F[Unit] =
+    log(KernelLogLevel.Trace, record)
+  final def logDebug(record: Builder => Builder): F[Unit] =
+    log(KernelLogLevel.Debug, record)
+  final def logInfo(record: Builder => Builder): F[Unit] =
+    log(KernelLogLevel.Info, record)
+  final def logWarn(record: Builder => Builder): F[Unit] =
+    log(KernelLogLevel.Warn, record)
+  final def logError(record: Builder => Builder): F[Unit] =
+    log(KernelLogLevel.Error, record)
```
morgen-peschke, line 50: Here's what the inlined version looks like:

```suggestion
  def log(level: KernelLogLevel, record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit]

  final def logTrace(record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] =
    log(KernelLogLevel.Trace, record)
  final def logDebug(record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] =
    log(KernelLogLevel.Debug, record)
  final def logInfo(record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] =
    log(KernelLogLevel.Info, record)
  final def logWarn(record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] =
    log(KernelLogLevel.Warn, record)
  final def logError(record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] =
    log(KernelLogLevel.Error, record)
```
The readability is about the same, and it means that `NoOpKernelLogger` can be implemented like this:
```scala
object NoOpLoggerKernel {
  def apply[F[_]: Applicative, Ctx]: LoggerKernel[F, Ctx] = (level, record) => Applicative[F].unit
}
```

core/shared/src/main/scala/org/typelevel/log4cats/SamLogger.scala
```diff
@@ -0,0 +1,130 @@
...
+import cats.*
+import cats.data.{EitherT, Kleisli, OptionT}
+
+/**
+ * A SAM-based Logger that extends LoggerKernel and provides a user-friendly interface. This is the
+ * new design that will eventually replace the current Logger trait.
+ */
+abstract class SamLogger[F[_], Ctx] extends LoggerKernel[F, Ctx] {
+
+  final def info(logBit: LogRecord[Ctx], others: LogRecord[Ctx]*)(implicit
```
morgen-peschke, line 28: I would _strongly_ recommend against this API design. 

It's shiny, but very unfriendly to users, because it would allow stuff like this:

```
logger.info(caughtException, otherCaughtException)
```

This compiles and emits an `Info`-level log without a message, and only one of the two exceptions that were intended to be logged.

core/shared/src/main/scala/org/typelevel/log4cats/SamLogger.scala
```diff
@@ -0,0 +1,130 @@
...
+package org.typelevel.log4cats
+
+import cats.*
+import cats.data.{EitherT, Kleisli, OptionT}
+
+/**
+ * A SAM-based Logger that extends LoggerKernel and provides a user-friendly interface. This is the
+ * new design that will eventually replace the current Logger trait.
+ */
+abstract class SamLogger[F[_], Ctx] extends LoggerKernel[F, Ctx] {
```
morgen-peschke, line 26: `LoggerKernel` should probably not be exposed to API users. 

`SamStructuredLogger` has the right idea - expose a rich and helpful API to the user, while allowing implementers to override a single field (`kernel`) to create their own variations.

core/shared/src/main/scala/org/typelevel/log4cats/SamLoggerAdapter.scala
```diff
@@ -0,0 +1,159 @@
...
+ *     support structured logging. Only message and throwable information is preserved.
+ *   - **Performance**: The `toLoggerKernel` method creates a full log record to extract message and
+ *     throwable, which has some overhead.
+ *
+ * ## Best Practices
+ *
+ *   - Use `toLogger` for backward compatibility when migrating to new backends
+ *   - Use `loggerToSamLogger` when you need structured logging capabilities
+ *   - Avoid round-trip conversions (LoggerKernel → Logger → LoggerKernel) as they lose context
+ */
```
morgen-peschke, line 57: As much as I'd love for this to be the case, ScalaDoc doesn't use Markdown syntax, here's how this would look in (mostly) ScalaDoc's wiki syntax (I can't always get the list syntax to work, so I have to fall back to `<ul>` or `<ol>` tags)

```suggestion
/**
 * Adapter layer to bridge the new SAM LoggerKernel with the existing Logger interface. This allows
 * gradual migration from the old multi-method design to the new SAM design.
 *
 * = Usage =
 * == Converting SAM LoggerKernel to old Logger interface ==
 * {{{
 * val kernel = Slf4jLoggerKernel.fromName[IO]("MyApp")
 * val oldLogger = SamLoggerAdapter.toLogger(kernel)
 *
 * // Use old API
 * oldLogger.info("Works with old API")
 * oldLogger.error(exception)("Works with old API")
 * }}}
 *
 * == Converting old Logger to SAM LoggerKernel ==
 * {{{
 * val oldLogger  Slf4jLogger.getLogger[IO]
 * val samLogger = SamLoggerAdapter.loggerToSamLogger(oldLogger)
 *
 * // Use new SAM API
 * samLogger.info("Works with new API", ("key", "value"))
 * }}}
 *
 * == Limitations ==
 *
 * <ul>
 *
 *   <li>
 *     '''Context Loss''': When converting from LoggerKernel → Logger → LoggerKernel, structured
 *     context information (key-value pairs) is lost because the old Logger interface doesn't
 *     support structured logging. Only message and throwable information is preserved.
 *   </li>
 *   
 *   <li>
 *     '''Performance''': The `toLoggerKernel` method creates a full log record to extract message and
 *     throwable, which has some overhead.
 *   </li>
 *
 * </ul>
 * 
 * == Best Practices ==
 *
 * <ul>
 *
 *   <li>Use `toLogger` for backward compatibility when migrating to new backends</li>
 *
 *   <li>Use `loggerToSamLogger` when you need structured logging capabilities</li>
 *
 *   <li>Avoid round-trip conversions (LoggerKernel → Logger → LoggerKernel) as they lose context</li>
 *
 * </ul>
 */
```

core/shared/src/main/scala/org/typelevel/log4cats/SamLoggerAdapter.scala
```diff
@@ -0,0 +1,159 @@
...
+object SamLoggerAdapter {
```
morgen-peschke, line 58: Instead of wrapping them, you can rewrite the existing implementations in terms of `LoggerKernel` and the need for this conversion goes away.

Code that is written against the old interfaces works just the same as ever, with the exceptions of a few middleware classes that need to be simplified.

core/shared/src/main/scala/org/typelevel/log4cats/SamStructuredLogger.scala
```diff
@@ -0,0 +1,266 @@
...
+
+/**
+ * A SAM-based StructuredLogger that extends Logger and provides structured logging capabilities.
+ * This implementation uses the new SAM LoggerKernel design for better performance and middleware
+ * compatibility.
+ */
+trait SamStructuredLogger[F[_]] extends Logger[F] {
+  protected def kernel: LoggerKernel[F, String]
+
+  def trace(ctx: Map[String, String])(msg: => String): F[Unit] =
```
morgen-peschke, line 31: We should be able to make these final, as they delegate to `kernel`, which should be the overwritten field.

core/shared/src/main/scala/org/typelevel/log4cats/SamStructuredLogger.scala
```diff
@@ -0,0 +1,266 @@
...
+
+  implicit def eitherTSamStructuredLogger[F[_]: Functor, E](implicit
+      ev: SamStructuredLogger[F]
+  ): SamStructuredLogger[EitherT[F, E, *]] =
+    ev.mapK(EitherT.liftK[F, E])
+
+  implicit def kleisliSamStructuredLogger[F[_], A](implicit
+      ev: SamStructuredLogger[F]
+  ): SamStructuredLogger[Kleisli[F, A, *]] =
+    ev.mapK(Kleisli.liftK[F, A])
```
morgen-peschke, line 265: We _really_ don't want to add implicit conversions for this sort of thing. It leads to really confusing code.

core/shared/src/main/scala/org/typelevel/log4cats/ConsoleLoggerKernel.scala
```diff
@@ -0,0 +1,57 @@
...
+ */
+class ConsoleLoggerKernel[F[_], Ctx](implicit F: Sync[F]) extends LoggerKernel[F, Ctx] {
+
+  def log(level: KernelLogLevel, record: Log.Builder[Ctx] => Log.Builder[Ctx]): F[Unit] = {
+    F.delay {
+      val logRecord = record(Log.mutableBuilder[Ctx]()).build()
+
+      val timestamp = logRecord.timestamp.map(_.toMillis).getOrElse(System.currentTimeMillis())
+      // Use simple timestamp formatting instead of java.time.Instant for Scala Native compatibility
+      val timeStr = s"${new java.util.Date(timestamp).toString}"
```
morgen-peschke, line 32: cleanup: Scala Native should support [Instant](https://scala-native.org/en/stable/lib/javalib.html#java-time), and it's better to avoid the `java.util` time classes whenever possible.

core/shared/src/test/scala/org/typelevel/log4cats/NoOpLoggerKernelTest.scala
```diff
@@ -0,0 +1,41 @@
...
+import org.typelevel.log4cats._
+
+class NoOpLoggerKernelTest extends CatsEffectSuite {
+  test("NoOpLoggerKernel should do nothing and not fail") {
+    val kernel = NoOpLoggerKernel[IO, String]
+    val logger = SamLogger.wrap(kernel)
+    // All of these should be no-ops and not throw
+    logger.info("This should not appear").void *>
+      logger.error("This should not appear").void *>
+      logger.warn("This should not appear").void *>
```
morgen-peschke, line 30: cleanup: `logger.warn` and friends already return `F[Unit]`, so you don't need to add `.void`

core/shared/src/test/scala/org/typelevel/log4cats/SamStructuredLoggerTest.scala
```diff
@@ -0,0 +1,51 @@
...
+
+import cats.effect.IO
+import munit.CatsEffectSuite
+
+class SamStructuredLoggerTest extends CatsEffectSuite {
+
+  test("SamStructuredLogger should handle context with multiple entries") {
+    var capturedLogs: List[(KernelLogLevel, Log.Builder[String] => Log.Builder[String])] = Nil
+
+    val testKernel = new LoggerKernel[IO, String] {
```
morgen-peschke, line 27: issue: Instead of inlining this with a `var`, it'd be better to rework the existing test loggers to use `LoggerKernel`

core/shared/src/main/scala/org/typelevel/log4cats/SamLoggerAdapter.scala
Jay-Lokhande: That makes sense. I think removing the SamLoggerAdapter and making the necessary changes in the old interfaces would simplify things.

build.sbt
```diff
@@ -52,7 +52,34 @@ lazy val core = crossProject(JSPlatform, JVMPlatform, NativePlatform)
...
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("org.typelevel.log4cats.Logger.kernel"),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "org.typelevel.log4cats.extras.DeferredLogger.kernel"
+      ),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "org.typelevel.log4cats.extras.DeferredSelfAwareStructuredLogger.kernel"
+      ),
+      ProblemFilters.exclude[ReversedMissingMethodProblem](
+        "org.typelevel.log4cats.extras.DeferredStructuredLogger.kernel"
+      )
```
morgen-peschke: Bumping `tlBaseVersion` will tell MiMa that you're going to do a minor version bump, and these will go away. 

```
ThisBuild / tlBaseVersion := "2.8"
```

core/shared/src/main/scala/org/typelevel/log4cats/SamStructuredLogger.scala
```diff
@@ -0,0 +1,139 @@
...
+
+import cats.*
+import cats.Show.Shown
+
+/**
+ * A SAM-based StructuredLogger that extends Logger and provides structured logging capabilities.
+ * This implementation uses the new SAM LoggerKernel design for better performance and middleware
+ * compatibility.
+ */
+trait SamStructuredLogger[F[_]] extends Logger[F] {
```
morgen-peschke: cleanup: compare with `StructuredLogger` and you'll see they're basically identical (especially their APIs) so this one can be deleted

core/shared/src/main/scala/org/typelevel/log4cats/noop/NoOpLogger.scala
```diff
@@ -31,80 +31,29 @@ object NoOpLogger {
     if (evaluateArgs) strictImpl else lazyImpl
 
   private def lazyImpl[F[_]: Applicative] = new SelfAwareStructuredLogger[F] {
+    protected def kernel: LoggerKernel[F, String] = new LoggerKernel[F, String] {
+      def log(level: KernelLogLevel, record: Log.Builder[String] => Log.Builder[String]): F[Unit] =
+        Applicative[F].pure(())
+    }
```
morgen-peschke: cleanup: use the implementation you already have
```suggestion
    protected def kernel: LoggerKernel[F, String] = NoOpLoggerKernel[F, String]
```

iRevive (commented):
Looks good! Please take my feedback with a grain of salt, as I may be missing some important context. These are mainly points for discussion rather than suggested actions. 🙂

PR #924, opened 2025-08-03
Update Scala Native to 0.5.8

iRevive:


danicheg (approved):
Basically, lgtm. However, this requires broader agreement on how we plan to land this in http4s, for example. It's unclear how many CE RCs there will be before 3.7, which could make things quite cumbersome.

rossabaker:
I merged one of these in jawn-fs2.  It hasn't needed a release since 2022, and keeping it on a branch seemed like overkill.  I'm less sure what to do with this one.

iRevive:
CE 3.6 isn't available for Scala Native 0.5

